Issue #19, opened 2026-02-19
Extra DB Queries


Alikznollet:
- query for fetching if date is inbetween 2 others
- query for fetching if date is before value
- query for fetching if date is after value

PR #78, opened 2026-02-24
implement extra query filters on date before, after and inbetween

landgeet:
## 🎯 MAIN GOAL
- [x] deze PR implementeert een nieuwe feature
* implementeerd volgende filters in de event en de production search:
  * query for fetching if date is inbetween start and end date
  * query for fetching if date is before start and end date
  * query for fetching if date is after value start and end date
  
- [ ] deze PR fixt één of meerdere bugs
* nvt

## 🛠️ TESTING
Geen testen nodig (denk ik)

## 📝 DOCUMENTATION
- [x] Er is voldoende documentatie geschreven voor deze code

**Nog missende documentatie:**
* Extra uitleg in boven functie over argumenten

## 🔍 HOW TO TEST
Query in de database uittesten.

## ⚠️ REMAINING ISSUES
- [ ] Er bevinden zich nog onopgeloste bugs in de code
* nvt
- [ ] Deze code vereist dat er ergens anders nog iets aangepast wordt (bv. een aangepaste database structuur, API-endpoint, ...)
* nvt
 
## ✅ CLOSED ISSUES
- #19

### DIFF
--- a/backend/src/database/db.event.service.ts
+++ b/backend/src/database/db.event.service.ts
@@ -41,7 +41,7 @@ export class EventDatabaseService {
   /**
    * Generic get function for events.
    * @param filters gives the freedom to define the filters of the search you want.
-   * All filters are filtered by equals.
+   * All filters are filtered by equals except for date filters (see function).
    * Not all filters need to be defined, only the ones you want to use.
    * i.e: getEvents({p_id: id}) will give a list of all events with the given p_id.
    * @returns All events for the given filters.
@@ -50,6 +50,12 @@ export class EventDatabaseService {
     filters: Partial<{
       genre: string;
       date: string;
+      // Return events where the provided date lies between starttime and endtime
+      date_between: string;
+      // Return events where starttime is before the provided date
+      date_before: string;
+      // Return events where endtime is after the provided date
+      date_after: string;
       hall: string;
       id: number;
       production_id: number;
@@ -71,6 +77,28 @@ export class EventDatabaseService {
     if (filters.date) {
       conditions.push(`DATE(e.starttime) = $${i} OR DATE(e.endtime) = $${i}`);
       values.push(filters.date);
+      i++;
+    }
+
+    // Filter by given date lying between starttime and endtime (inclusive)
+    if (filters.date_between) {
+      // Use explicit timestamp comparison to include time component
+      conditions.push(`$${i}::timestamp BETWEEN e.starttime AND e.endtime`);
+      values.push(filters.date_between);
+      i++;
+    }
+
+    // Filter events whose starttime is before the provided date
+    if (filters.date_before) {
+      conditions.push(`e.starttime < $${i}::timestamp OR e.endtime < $${i}::timestamp`);
+      values.push(filters.date_before);
+      i++;
+    }
+
+    // Filter events whose endtime is after the provided date
+    if (filters.date_after) {
+      conditions.push(`e.endtime > $${i}::timestamp OR e.starttime > $${i}::timestamp`);
+      values.push(filters.date_after);
       i++;
     }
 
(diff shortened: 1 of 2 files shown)


Alikznollet (commented):
Ziet er wel goed uit denk ik, weet gewoon niet goed hoe ik moet verifiëren of dit werkt op het moment, maar dat is een probleem voor later I suppose?

landgeet:
> Ziet er wel goed uit denk ik, weet gewoon niet goed hoe ik moet verifiëren of dit werkt op het moment, maar dat is een probleem voor later I suppose?

Opzich kan je manueel de database opstarten en de letterlijke queries testen op de huidige data, maar ik denk dat dit vooral zal moeten gebeuren in #40

backend/src/database/db.event.service.ts
```diff
@@ -74,6 +80,28 @@ export class EventDatabaseService {
...
+    if (filters.date_between) {
+      // Use explicit timestamp comparison to include time component
+      conditions.push(`$${i}::timestamp BETWEEN e.starttime AND e.endtime`);
+      values.push(filters.date_between);
+      i++;
+    }
+
+    // Filter events whose starttime is before the provided date
+    if (filters.date_before) {
+      conditions.push(`e.starttime < $${i}::timestamp OR e.endtime < $${i}::timestamp`);
```
MrLegitFTW: this should work too if you do `e.starttime < $${i}::timestamp` therfore the second part is not necessary?

backend/src/database/db.event.service.ts
```diff
@@ -74,6 +80,28 @@ export class EventDatabaseService {
...
+      values.push(filters.date_before);
+      i++;
+    }
+
+    // Filter events whose endtime is after the provided date
+    if (filters.date_after) {
+      conditions.push(`e.endtime > $${i}::timestamp OR e.starttime > $${i}::timestamp`);
```
MrLegitFTW: same here but the other way around.

backend/src/database/db.production.service.ts
```diff
@@ -95,6 +101,28 @@ export class ProductionDatabaseService {
...
+    if (filters.date_between) {
+      // Use explicit timestamp comparison to include time component
+      conditions.push(`$${i}::timestamp BETWEEN e.starttime AND e.endtime`);
+      values.push(filters.date_between);
+      i++;
+    }
+
+    // Filter events whose starttime is before the provided date
+    if (filters.date_before) {
+      conditions.push(`e.starttime < $${i}::timestamp OR e.endtime < $${i}::timestamp`);
```
MrLegitFTW: same as in db.event.service

MrLegitFTW (commented):
This looks good only a few questions as seen below.

backend/src/database/db.event.service.ts
landgeet: Yes, good catch, this is no necessary indeed. I got confused with times and stuff. I will fix this.

landgeet:
This should be fixed

MrLegitFTW (approved):
This looks good. Just like @Alikznollet said we will need to wait for #40 to be sure this works as intended but I think this should be good. Do we need to update/add more GET's after this? If so you can create a new issue when merging this pr.

landgeet:
> This looks good. Just like @Alikznollet said we will need to wait for #40 to be sure this works as intended but I think this should be good. Do we need to update/add more GET's after this? If so you can create a new issue when merging this pr.

I don't think so because this is another filter and shouldn't really be an endpoint in the API.
We could make some functions that use these filters in (only) the services if we really want to, but maybe we discuss this on friday whether this is necessary or not.

MrLegitFTW:
> > This looks good. Just like @Alikznollet said we will need to wait for #40 to be sure this works as intended but I think this should be good. Do we need to update/add more GET's after this? If so you can create a new issue when merging this pr.
> 
> I don't think so because this is another filter and shouldn't really be an endpoint in the API. We could make some functions that use these filters in (only) the services if we really want to, but maybe we discuss this on friday whether this is necessary or not.

Sounds good to me

PR #102, opened 2026-02-27
E2e tests

Robindr145:
## 🎯 MAIN GOAL
- [ ] this PR implements a new feature
* describe the newly implemented features
* describe the fixed bugs
- [x] this PR implements some tests

## 🛠️ TESTING
- [x] all tests succeed

**Missing tests:**
* clarify if certain edge-cases still need testing

There is a test at line `~175` where, in normal behaviour, it has to generate an own generated error and it will display it in the terminal. If we don't want this error to be shown in the terminal, we can remove the comment at line `~117`. 

## 📝 DOCUMENTATION
- [x]  there is enough documentation written for this code

**Missing documentation:**

## 🔍 HOW TO TEST
1. Go to `/backend` in the terminal
2. type `npm run test:e2e`
4. Expected result: 
    test Suites: 1 passes, 1 total
    tests: 97 passed, 97 total

## ⚠️ REMAINING ISSUES
* explain where the bugs are located
* clarify what needs to be changed
already merged my branch with the hotfixes in `/backend/src/tag/tagcontroller.ts`

## ✅ CLOSED ISSUES
- #58


landgeet (commented):
The tests look really good and all pass.
The only problem is that these tests don't run when using the `npm run test` command.
Maybe this is intended, but this just makes it so you need to use two seperate commands to run all tests.

A possible solution could be to move the file into the `src/` directory and make it a real `.spec.ts` file so it would run.
Another solution could be that we just move all existing tests into the `test/` directory.

This is something we should discuss with the rest.

Alikznollet:
> The tests look really good and all pass. The only problem is that these tests don't run when using the `npm run test` command. Maybe this is intended, but this just makes it so you need to use two seperate commands to run all tests.
> 
> A possible solution could be to move the file into the `src/` directory and make it a real `.spec.ts` file so it would run. Another solution could be that we just move all existing tests into the `test/` directory.
> 
> This is something we should discuss with the rest.

It should be possible to add this directory to the "to run" tests in the `jest.config.ts`. Don't know how off the top of my head but should be a feature. No need to move any files then.

landgeet:
> > The tests look really good and all pass. The only problem is that these tests don't run when using the `npm run test` command. Maybe this is intended, but this just makes it so you need to use two seperate commands to run all tests.
> > A possible solution could be to move the file into the `src/` directory and make it a real `.spec.ts` file so it would run. Another solution could be that we just move all existing tests into the `test/` directory.
> > This is something we should discuss with the rest.
> 
> It should be possible to add this directory to the "to run" tests in the `jest.config.ts`. Don't know how off the top of my head but should be a feature. No need to move any files then.

Oh alright then, this would be nice to have.

Alikznollet (approved):

landgeet (approved):

PR #117, opened 2026-03-02
Insert Productions

### LINKED ISSUE #109: Insert production with existing ID
Implement a function `insertProduction` at endpoint `production/insert` that inserts a production that has a predefined ID.
This is necessary to insert the production data from CSV files into the database.
(not necessary for events, because these don't have a predefined ID)

Alikznollet:
## 🎯 MAIN GOAL
* Adds an endpoint that allows for inserting productions (forcibly). This will override any already existing production with the ID specified in the new one.

## 🛠️ TESTING
- [x] all tests succeed

**Missing tests:**
* Still need to write tests for this code.

## 📝 DOCUMENTATION

**Missing documentation:**
* Documentation is already there.

## 🔍 HOW TO TEST

Just try it out via the Swagger UI. It should be able to add Productions that don't already exist or overwrite existing ones.

## ✅ CLOSED ISSUES
- Closes #109

### DIFF
--- a/backend/src/database/db.production.service.ts
+++ b/backend/src/database/db.production.service.ts
@@ -279,6 +279,64 @@ export class ProductionDatabaseService {
   }
 
   /**
+   * Forcibly insert a production with an already existing ID into the Database.
+   * Will either create a new production or override an existing one.
+   * @param production The production object that we want to insert.
+   * @returns That same production object but returned from the Database.
+   */
+  async insertProduction(
+    production: ProductionDto
+  ): Promise<ProductionDto> {
+
+    const query = `
+      INSERT INTO productions (
+        id,
+        titel,
+        ondertitel,
+        description1,
+        description2,
+        genre,
+        planning_id
+      )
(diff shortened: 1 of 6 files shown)


backend/src/database/db.production.service.ts
```diff
@@ -278,6 +278,64 @@ export class ProductionDatabaseService {
     return result[0];
   }
 
+  /**
+   * Forcibly insert a production with an already existing ID into the Database.
```
MrLegitFTW: Maybe add here to clarify that this is the unsafe version of createProduction so it is clear that there are risks involved when calling this function.

MrLegitFTW (changes requested):
Looks good, maybe add a bit of clarification as to the dangers of using insertProduction.

MrLegitFTW (approved):

landgeet (commented):
Looks good, can be merged

landgeet (approved):

PR #118, opened 2026-03-02
Feature/genre to tag parsing

### LINKED ISSUE #116: Genre to Tag parsing.
- [ ] CSV Parser should parse every unique `genre` to a Tag-Object that then can be added to the database.
- [ ] Those tags then need to be linked to the production you're parsing.

This means you'll have to keep track of which tags are supposed to be linked with which productions, so you can perform those links after everything has been added.

landgeet:
## 🎯 MAIN GOAL
* makes production parser return tags and tags linked to production id
* implement `insertProductionsFromCSV` function that parses productions from a CSV file and inserts them into the database, along with their associated tags. 
* NEW: makes event parser also return location 
* NEW: tweak parsing and inserting logic for event to also create and link location to events

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Validate that the spec tests pass
2. Run both the scripts below:
```bash
cd backend
# runs event parsing
npx ts-node src/csv_parsing/run_parse_events.ts ../common/res/events_voorstellingen.csv 5
# runs production parsing
npx ts-node src/csv_parsing/run_parse_productions.ts ../common/res/productions_output.csv 5
```
Needed info about these scripts:
- these are temporary and are just for easier testing purposes (will be delete before merging)
- change the `5` and the end to a different number `n` to get `n` records of the parsed data printed to the terminal 
- when an invalid row is encountered or an error occurs, it will print a warning on the console and skip the row
- for productions it also prints all disctinct tags (genres) found in the file.
- NEW: for events it also prints all disctinct locations (halls) found in the file.
3. Validate that the ouput records corresponds to the structure of the csv data.

## ⚠️ REMAINING ISSUES
- should be none

## ✅ CLOSED ISSUES
- Closes #116


Alikznollet (approved):
Looks good to me!

MrLegitFTW (approved):
Looks good.

MatthieuDC (approved):
Looks good! We have to make sure that later on in the front-end, the timezones have to be translated to match the region

Alikznollet (approved):
Looks good, can be merged together with csv parser. We might have to make new issues then related to what we decide tomorrow about the database schemes.

Robindr145 (approved):
looks good

PR #135, opened 2026-03-04
Error codes

### LINKED ISSUE #130: Error codes
- use error code 401 if user is not autherized for something
- use error code 403 if it is forbidden (authed but not allowed)
- use error code 410 if the result of a get etc is nothing (not found)
- use error code 404 if an ENDPOINT is incorrect (does not exist)
- use error code 418 as easter egg (look it up)

MatthieuDC:
## 🎯 MAIN GOAL
* different types of error codes based on how they're thrown

## 🛠️ TESTING
- [x] all tests succeed

**Missing tests:**
* tests still have to be written (TODO)

## 📝 DOCUMENTATION

**Missing documentation:**
* maybe extra documentation that notes how an error is thrown in the different classes?

## 🔍 HOW TO TEST
1. Start server with npm run start:dev (in /backend)
2. In a separate console tab, use curl commands to simulate the different errors
3. Watch the error codes

## ✅ CLOSED ISSUES
- #130

### DIFF
--- a/backend/src/app.controller.ts
+++ b/backend/src/app.controller.ts
@@ -7,7 +7,12 @@ export class AppController {
   constructor(private readonly appService: AppService) {}
 
   @Get()
-  @Redirect("/docs", 301)
+  @Redirect()
+  redirectToDocs() {
+    const isProduction = process.env.NODE_ENV === "production";
+    return { url: isProduction ? "/api/docs" : "/docs", statusCode: 301 };
+  }
+
   root(): void {}
 
   // Easter egg 418
--- a/backend/src/app.controller.ts
+++ b/backend/src/app.controller.ts
@@ -1,5 +1,6 @@
 import { Controller, Get, Redirect } from "@nestjs/common";
 import { AppService } from "./app.service";
+import { TeapotException } from "./common/exceptions";
 
 @Controller()
 export class AppController {
@@ -8,4 +9,10 @@ export class AppController {
   @Get()
   @Redirect("/docs", 301)
   root(): void {}
+
+  // Easter egg 418
+  @Get("coffee")
+  makeCoffee() {
+    throw new TeapotException();
+  }
 }
--- a/backend/src/blog/blog.controller.ts
+++ b/backend/src/blog/blog.controller.ts
@@ -13,7 +13,7 @@ import {
 import { BlogService } from "./blog.service";
 import { BlogSchema, CreateBlogSchema, UpdateBlogSchema } from "@repo/common";
 import { ZodValidationPipe } from "../common/pipes/zod.validation.pipe";
-import { ResourceGoneException } from "src/common/exceptions";
+import { ResourceGoneException } from "../common/exceptions";
 import { BlogDto, CreateBlogDto, UpdateBlogDto } from "../dto/dto";
 import {
   ApiBody,
--- a/backend/src/blog/blog.controller.ts
+++ b/backend/src/blog/blog.controller.ts
@@ -13,6 +13,7 @@ import {
 import { BlogService } from "./blog.service";
 import { BlogSchema, CreateBlogSchema, UpdateBlogSchema } from "@repo/common";
 import { ZodValidationPipe } from "../common/pipes/zod.validation.pipe";
+import { ResourceGoneException } from "src/common/exceptions";
 import { BlogDto, CreateBlogDto, UpdateBlogDto } from "../dto/dto";
 import {
   ApiBody,
@@ -52,7 +53,11 @@ export class BlogController {
   async getBlogById(
     @Param("blogId", ParseIntPipe) blogId: number,
   ): Promise<BlogDto> {
-    return await this.blogService.getBlogById(blogId);
+    const blog = await this.blogService.getBlogById(blogId);
+    if (!blog) {
+      throw new ResourceGoneException(`Blog ${blogId} not found.`);
+    }
+    return blog;
   }
 
   /**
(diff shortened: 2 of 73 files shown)


MatthieuDC:
pls check if test coverage is ok

backend/src/app.controller.ts
```diff
@@ -8,4 +9,10 @@ export class AppController {
   @Get()
   @Redirect("/docs", 301)
   root(): void {}
+
+  // Easter egg 418
+  @Get("coffee")
+  makeCoffee() {
+    throw new TeapotException();
+  }
```
Alikznollet, line 22: This is fun haha.

Alikznollet (changes requested):
In general the exception handling looks pretty good, the only thing I'd say is that it could be that for most cases an exception would be thrown way deeper into the services.

Take for example getting a production by ID. This will query the DB in the ProductionDatabaseService, but this will also throw an exception in there, which will then just not continue execution and so the branch will never reach your exception code.

I would rewrite this to handle the exceptions at query level instead of inside the controller. But maybe we need @MrLegitFTW opinion on this too.

Update: Oh and coverage looks fine by the way.

backend/src/blog/blog.controller.ts
```diff
@@ -88,7 +93,11 @@ export class BlogController {
     @Param("blogId", ParseIntPipe) blogId: number,
     @Body(new ZodValidationPipe(BlogSchema)) blog: BlogDto,
   ): Promise<BlogDto> {
-    return await this.blogService.replaceBlog(blogId, blog);
+    const updated = await this.blogService.replaceBlog(blogId, blog);
```
MrLegitFTW: I will second @Alikznollet on this. throwing the ResourceGone over here would work, however at the moment the db service checks in the updateBlog function if the database finds a blog with that id and if not already throws an exception, therfore I would suggest moving that exception te there as it reflect the responsibility of that piece of code to return a blog or throw an error if it cannot.

This is in the updateBlog-function in db.blog.service:
```suggestion
        const result = await this.db.query<BlogDto>(query, values);

    if (result.length === 0) {
      throw new ResourceGoneError("Blog not found"); // throw the error over here
    }

    return result[0];
```

backend/src/blog/blog.controller.ts
```diff
@@ -123,6 +136,10 @@ export class BlogController {
   async deleteBlog(
     @Param("blogId", ParseIntPipe) blogId: number,
   ): Promise<void> {
+    const blog = await this.blogService.getBlogById(blogId);
```
MrLegitFTW: the delete function does not return anything right now for silent handling. I would not like to send 2 query's in order to determine if an error should be thrown instead I would suggest to either throw the error in the delete function in the db service or keep it silent as we have it now. both work and could be up for debate?

For instance in the db.blog.service deleteBlog function we could make:
```suggestion
    async deleteBlog(id: number): Promise<void> {
      const query = `DELETE FROM blogs WHERE id = $1 RETURNING id`; // you would need to return the id for this to work.

      // DELETE returns the deleted amount (should be 1):
      const result = await this.db.query(query, [id]);
    
      if (result.length === 0){
        throw new ResourceGoneException('') // throw it here?
      }
  }
```

backend/src/common/exceptions.ts
```diff
@@ -0,0 +1,15 @@
...
+
+// 410 Gone for resources not found
+export class ResourceGoneException extends HttpException {
+    constructor(message: string = 'Resource is gone') {
+        super(message, HttpStatus.GONE);
+    }
+}
+
+// 418 I'm a teapot easter egg
+export class TeapotException extends HttpException {
```
MrLegitFTW, line 11: :-)

backend/src/blog/blog.controller.ts
```diff
@@ -52,7 +53,11 @@ export class BlogController {
   async getBlogById(
     @Param("blogId", ParseIntPipe) blogId: number,
   ): Promise<BlogDto> {
-    return await this.blogService.getBlogById(blogId);
+    const blog = await this.blogService.getBlogById(blogId);
+    if (!blog) {
```
MrLegitFTW: I would suggest moving this to the db service as an error is already thrown there.

Here is an example in the db service getById functions:

```suggestion
    const result = await this.db.query<BlogDto>(query, [id]);

    if (result.length === 0) {
      throw new Error("Blog not found"); // change the error here.
    }
    return result[0];
```

MrLegitFTW (changes requested):
I think most ResourceGoneExceptions could be moved to the db services as I have highlighted below for the blogs. This applies to the other services. In the case of the delete functions it is up for debate and I will leave the choice to you if we keep the silent handling as we do now (no errors) or if you want to change it to the example I gave. In any case I am happy with the structure of the errors and easter egg :-)

MatthieuDC:
Database changes have to be updated first before this can be finished, since some tests fail because of the mismatch on develop branch

backend/src/database/db.blog.service.ts
```diff
@@ -190,10 +179,12 @@ export class BlogDatabaseService {
    * (silent handling)
```
MrLegitFTW: Very nitpicky I know. Perhaps remove the silent handling comment now seeing as it is no longer silent. And remove the no errors are thrown part.

MrLegitFTW (changes requested):
This looks strong. Need to be the nitpicker sadly. Could you update the comments on the delete-functions to reflect they are no longer silent.

Alikznollet (commented):
Looks good to me, maybe just apply what Sebastien said and this will be ready to merge.

Issue #146, opened 2026-03-05
Price DB Service.


Alikznollet:
- [ ] Create a DB Service for Price objects.

PR #157, opened 2026-03-05
Db update

### LINKED ISSUE #148: Price DB Schema.
- [ ] Add new Schemas for Price in DB and common.

MrLegitFTW:
## 🎯 MAIN GOAL
* updates the price in the events to be a seperate table.

## 🛠️ TESTING
- [x] all tests succeed

**Missing tests:**
* @landgeet please check if the csv is still good.

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Go to page X / Run command Y
2. Press Z
3. Expected result: ...

## ⚠️ REMAINING ISSUES
* see #146 

image

## ✅ CLOSED ISSUES
- #148

### DIFF
--- a/backend/src/csv_parsing/csv_file_parser.spec.ts
+++ b/backend/src/csv_parsing/csv_file_parser.spec.ts
@@ -85,7 +85,6 @@ describe("CSVFileParser.transformEventRow", () => {
       Endtime: "2024-01-01 12:00:00",
       Production: "5",
       Genre: "drama",
-      Price: "25",
       Hall: "Main Hall",
     };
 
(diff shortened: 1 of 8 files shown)


Alikznollet (approved):
Looks good!

backend/src/csv_parsing/csv_file_parser.spec.ts
landgeet, line 1: You don't need to remove the Price fields here (price with a lowercase letter is necessary to delete tho), because I will let the parser parse the price later even though it's not in their csv files. This gives the opportunity to parse a csv file that contains the prices if they would like to.
TLDR: I would let the Price fields (not price) stay there because i'll add them back later anyways

landgeet (commented):
Everything looks good, just a small remark regarding the Price fields in de tests.

landgeet (approved):

backend/src/csv_parsing/csv_file_parser.spec.ts
landgeet: Nevermind, i realised it would be better to let them input a seperate CSV for prices because an event has multiple prices with different names and the event CSV would get really messy. So ignore what i said.

PR #159, opened 2026-03-06
Communication between frontend and backend

Nowak-Dominik:
## 🎯 MAIN GOAL
* Communication between frontend and backend
* None

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* None

## 📝 DOCUMENTATION

**Missing documentation:**
* None

## 🔍 HOW TO TEST
* Run the test corresponding to the new features.

## ⚠️ REMAINING ISSUES
* No bugs found yet
* Nothing

## ✅ CLOSED ISSUES
- #139

### DIFF
--- a/frontend/app/composables/useApi.ts
+++ b/frontend/app/composables/useApi.ts
@@ -69,7 +69,6 @@ export function useApi() {
 
   /**
    * Fallback error handler used when no onError is passed to a request.
-   * Redirects to /login on 401/403.
    */
   function handleDefaultError(status: number, message: string): void {
     if (status === 401 || status === 403) {
--- a/frontend/app/composables/useApi.ts
+++ b/frontend/app/composables/useApi.ts
@@ -1,5 +1,3 @@
-import { useAuth } from "./useAuth";
-
 export interface ApiOptions<TBody = unknown> {
   body?: TBody;
   headers?: Record<string, string>;
--- a/frontend/app/composables/useApi.ts
+++ b/frontend/app/composables/useApi.ts
@@ -1,0 +1,126 @@
+import { useAuth } from "./useAuth";
+
+export interface ApiOptions<TBody = unknown> {
+  body?: TBody;
+  headers?: Record<string, string>;
+  /** Override the default error handling for this specific request. */
+  onError?: (status: number, message: string) => void;
+}
+
+export interface ApiResponse<TData> {
+  data: TData | null;
+  error: string | null;
+  status: number | null;
+}
+
+export function useApi() {
+  const config = useRuntimeConfig();
+  const baseUrl = config.public.apiBase as string;
+  const { apiKey } = useAuth();
+
+  /**
+   * Base fetch wrapper used by all method helpers.
+   * Attaches the x-api-key header when the user is logged in.
+   * On non-ok responses, calls onError if provided, otherwise falls back to handleDefaultError.
+   */
+  async function request<TData, TBody = unknown>(
+    method: "GET" | "POST" | "PUT" | "PATCH" | "DELETE",
+    endpoint: string,
+    options: ApiOptions<TBody> = {}
+  ): Promise<ApiResponse<TData>> {
+    const { body, headers: extraHeaders, onError } = options;
+
+    const headers: Record<string, string> = {
+      "Content-Type": "application/json",
+      ...(apiKey.value ? { "x-api-key": apiKey.value } : {}),
+      ...extraHeaders,
+    };
+
+    try {
+      const response = await fetch(`${baseUrl}${endpoint}`, {
+        method,
+        headers,
+        body: body !== undefined ? JSON.stringify(body) : undefined,
+      });
+
+      const text = await response.text();
+      const data: TData | null = text ? (JSON.parse(text) as TData) : null;
+
+      if (!response.ok) {
+        const message =
+          (data as { message?: string } | null)?.message ??
+          `Request failed with status ${response.status}`;
+
+        onError ? onError(response.status, message) : handleDefaultError(response.status, message);
+
+        return { data: null, error: message, status: response.status };
+      }
+
+      return { data, error: null, status: response.status };
+    } catch (err) {
+      const message =
+        err instanceof Error ? err.message : "An unexpected network error occurred.";
+
+      onError ? onError(0, message) : handleDefaultError(0, message);
+
+      return { data: null, error: message, status: null };
+    }
+  }
+
+  /**
+   * Fallback error handler used when no onError is passed to a request.
+   * Redirects to /login on 401/403.
+   */
+  function handleDefaultError(status: number, message: string): void {
+    if (status === 401 || status === 403) {
+      console.error(`[useApi] Auth error (${status}): ${message}`);
+      navigateTo("/login");  // TODO dit moet de admin subdomain login zijn
+    } else if (status === 404) {
+      console.warn(`[useApi] Not found (404): ${message}`);
+    } else if (status === 0) {
+      console.error(`[useApi] Network error: ${message}`);
+    } else {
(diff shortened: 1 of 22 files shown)


frontend/nuxt.config.ts
```diff
@@ -11,6 +11,8 @@ export default defineNuxtConfig({
   runtimeConfig: {
     public: {
       apiBase: process.env.NUXT_API_BASE || "http://localhost:3000",
+      adminApiBase: process.env.NUXT_API_BASE || "http://localhost:3000",
```
MrLegitFTW: I would think that we will be making a seperate website for this page? Therefore this will probably be moved over to that config file. However as discussed on discord this is fine here for now.

frontend/app/utils/routes.ts
```diff
@@ -0,0 +1,33 @@
+export const ROUTES = {
```
MrLegitFTW, line 8: This is good. I like the structure.

frontend/app/utils/apiRoutes.ts
```diff
@@ -0,0 +1,33 @@
+export const API_ROUTES = {
```
MrLegitFTW, line 7: This is good structure. Allows for simple addition later.

frontend/app/composables/useBlogApi.ts
```diff
@@ -0,0 +1,26 @@
+import type { Blog, CreateBlog, UpdateBlog } from "@repo/common";
+import { API_ROUTES } from "../utils/apiRoutes";
+
+export function useBlogApi() {
```
MrLegitFTW, line 19: This could use some documentation for example.

frontend/app/composables/useEventApi.ts
```diff
@@ -0,0 +1,56 @@
+import type {
+  Event,
+  CreateEvent,
+  UpdateEvent,
+  FilterEvent,
+  Location,
+} from "@repo/common";
+import { API_ROUTES } from "../utils/apiRoutes";
+
+export function useEventApi() {
```
MrLegitFTW, line 19: This too.

frontend/app/composables/useLocationApi.ts
```diff
@@ -0,0 +1,23 @@
+import type { Location, CreateLocation, UpdateLocation } from "@repo/common";
+import { API_ROUTES } from "../utils/apiRoutes";
+
+export function useLocationApi() {
```
MrLegitFTW, line 19: Here too etc... You get the idea.

MrLegitFTW (changes requested):
This is very solid. Big fan of the structure you used here only 2 remarks realy. Would like to see more documentation for the functions. I know they may be small functions but it realy does help. And secondly (sadly) we do need some tests to verify that all is in working order. Great job so far :-)

MrLegitFTW:
Forgot to mention that you will have to add language support to most request. You can do this after #186 and #187 are merged into dev. (see index.ts under common for the language definition)

Nowak-Dominik:
force-pushed after rebasing from develop to be up to date

Alikznollet (commented):
So I hate to be the bringer of bad news and this definitely looks nice but did you take a look at 
https://nuxt.com/docs/4.x/getting-started/data-fetching
before starting with this? Nuxt has built in data fetching that would I feel like result in a lot less code that needs to be maintained.

Nowak-Dominik:
> So I hate to be the bringer of bad news and this definitely looks nice but did you take a look at https://nuxt.com/docs/4.x/getting-started/data-fetching before starting with this? Nuxt has built in data fetching that would I feel like result in a lot less code that needs to be maintained.

Yeah, good call. I looked into it already before but you're right that we should use Nuxt's $fetch. I'll update useApi.ts to use it.

The reason I didn't use it from the start is that the main benefit of useFetch/useAsyncData is preventing double-fetching in SSR mode, but since we have ssr: false in nuxt.config.ts (because we have a separate NestJS backend), that advantage doesn't apply to us.

That said, switching to $fetch over native fetch still gives us some small but worthwhile improvements like: automatic JSON parsing, better built-in error handling, and interceptor support which lets us clean up how we attach the API key. Not a huge difference since useApi.ts already handles most of this manually, but it removes some boilerplate and aligns better with how Nuxt is meant to be used.

Alikznollet:
> > So I hate to be the bringer of bad news and this definitely looks nice but did you take a look at https://nuxt.com/docs/4.x/getting-started/data-fetching before starting with this? Nuxt has built in data fetching that would I feel like result in a lot less code that needs to be maintained.
> 
> Yeah, good call. I looked into it already before but you're right that we should use Nuxt's $fetch. I'll update useApi.ts to use it.
> 
> The reason I didn't use it from the start is that the main benefit of useFetch/useAsyncData is preventing double-fetching in SSR mode, but since we have ssr: false in nuxt.config.ts (because we have a separate NestJS backend), that advantage doesn't apply to us.
> 
> That said, switching to $fetch over native fetch still gives us some small but worthwhile improvements like: automatic JSON parsing, better built-in error handling, and interceptor support which lets us clean up how we attach the API key. Not a huge difference since useApi.ts already handles most of this manually, but it removes some boilerplate and aligns better with how Nuxt is meant to be used.

My goat, I'll look into this PR again later today!

frontend/nuxt.config.ts
MrLegitFTW: these admin urls can be removed now seeing as we will not be IP-locking anymore.

MrLegitFTW (changes requested):
This looks much cleaner now it uses Nuxt's fetching code wise. I will leave the reviewing of that to @Alikznollet however. From my side it looks good, just need to remove the adminURL in the config seeing as that is no longer needed.

frontend/nuxt.config.ts
Nowak-Dominik: ow yeah you are right, will do

frontend/app/composables/useApi.ts
```diff
@@ -0,0 +1,124 @@
...
+  function patch<TData, TBody = unknown>(
+    endpoint: string,
+    body: TBody,
+    options?: Omit<ApiOptions<TBody>, "body">
+  ): Promise<ApiResponse<TData>> {
+    return request<TData, TBody>("PATCH", endpoint, { ...options, body });
+  }
+
+  /** deletion — removes the object. */
+  function del<TData>(
```
landgeet, line 116: i would refactor this to `delete` instead of `del` just for extra clarity (i know this is insanely nitpicky haha)

frontend/app/utils/routes.ts
landgeet, line 8: This file could use some documentation. A general comment explaining what this is used for should be good enough.

landgeet (changes requested):
This looks really good, well done!
- Documentation make the code clear and easy to digest.
- All tests pass and cover the main functionality.
- Everything works (as far as i can tell).

There are a couple really small nitpicky remarks listed below. If these get resolved, i'll approve!

frontend/app/composables/useApi.ts
Nowak-Dominik: I can not do that:

Image

'delete' is actually a reserved keyword in JavaScript/TypeScript so it can't be used as a function name.

landgeet: Oh yeah i forgot about that, this is fine then.

frontend/app/utils/routes.ts
Nowak-Dominik: Consider it done!

landgeet (approved):
Everything is good now!

MrLegitFTW (approved):
After all these changes I am confident in this pr.

Alikznollet (approved):
This looks great. Nothing bad to say.

PR #160, opened 2026-03-06
Price db service

### LINKED ISSUE #146: Price DB Service.

landgeet:
## 🎯 MAIN GOAL
* implement price db service
* implement event db service price-related function

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
Test the queries on the database, same as always.

## ✅ CLOSED ISSUES
- closes #146

### DIFF
--- a/backend/src/database/db.price.service.ts
+++ b/backend/src/database/db.price.service.ts
@@ -29,7 +29,7 @@ export class PriceDatabaseService {
    * @param page page index (starts at 0)
    * @return prices
    */
-  async getPrices(amount: number, page: number): Promise<PriceDto[]> {
+  async getPrices(amount: number = 0, page: number = 0): Promise<PriceDto[]> {
     const offset = page * amount;
 
     if (amount === 0) {
--- a/backend/src/database/db.price.service.ts
+++ b/backend/src/database/db.price.service.ts
@@ -1,0 +1,156 @@
+import { BadRequestException, Injectable } from "@nestjs/common";
+import { DbService } from "./db.service";
+import { CreatePriceDto, PriceDto, UpdatePriceDto } from "../dto/dto";
+
+@Injectable()
+export class PriceDatabaseService {
+  // need to give a db service as param when used. -> see db.service.
+  constructor(private db: DbService) {}
+
+  /**
+   * Get a single price by their ID.
+   * @param id The ID we're trying to fetch.
+   * @returns The Price if there is one.
+   */
+  async getPriceById(id: number): Promise<PriceDto> {
+    const query = `SELECT * price FROM prices WHERE id = $1`;
+
+    const result = await this.db.query<PriceDto>(query, [id]);
+
+    if (result.length === 0) {
+      throw new Error("Price not found");
+    }
+    return result[0];
+  }
+
+  /**
+   * Get prices with pagination
+   * @param amount number of prices per page (if amount=0, it will default to grabbing all prices)
+   * @param page page index (starts at 0)
+   * @return prices
+   */
+  async getPrices(amount: number, page: number): Promise<PriceDto[]> {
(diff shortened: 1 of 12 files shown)


backend/src/database/db.price.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    return result[0];
+  }
+
+  /**
+   * Get prices with pagination
+   * @param amount number of prices per page (if amount=0, it will default to grabbing all prices)
+   * @param page page index (starts at 0)
+   * @return prices
+   */
+  async getPrices(amount: number, page: number): Promise<PriceDto[]> {
```
Alikznollet: Maybe add defaults of 0 for both here?

Alikznollet (changes requested):
One little nitpick about default values for parameters, for the rest this looks good!

Alikznollet:
I will be starting from this branch for my implementation of the price endpoints.

backend/src/database/db.price.service.ts
landgeet: Should be fixed

Alikznollet (approved):
Looks good.

backend/src/database/db.price.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      SELECT * 
+      FROM prices
+      ORDER BY id
+      `;
+
+      const result = await this.db.query<PriceDto>(query);
+
+      if (result.length === 0) {
+        throw new Error("no prices found");
+      }
```
Alikznollet: Can we maybe make it a habit of not erroring out here if we get an  empty result? Instead just return the empty list.

backend/src/database/db.price.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    LIMIT $1 OFFSET $2
+    `;
+
+    const result = await this.db.query<PriceDto>(query, [amount, offset]);
+
+    if (result.length === 0) {
+      throw new Error("no prices found");
+    }
+
+    return result;
```
Alikznollet: Same here, please return the empty list, it's a bit misleading to see an internal server error if there are just no prices.

Alikznollet (changes requested):
Small changes I saw when working on the endpoints.

Alikznollet:
@landgeet Oh and change the price field of PriceSchema to a float32 instead of a string.

backend/src/database/db.price.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  // need to give a db service as param when used. -> see db.service.
+  constructor(private db: DbService) {}
+
+  /**
+   * Get a single price by their ID.
+   * @param id The ID we're trying to fetch.
+   * @returns The Price if there is one.
+   */
+  async getPriceById(id: number): Promise<PriceDto> {
+    const query = `SELECT * price FROM prices WHERE id = $1`;
```
Alikznollet, line 16:
```suggestion
    const query = `SELECT * FROM prices WHERE id = $1`;
```

Feel like this was a typo.

common/src/database_objects.ts
```diff
@@ -72,6 +71,17 @@ export const FilterEventSchema = z.object({
   limit: z.coerce.number().min(1).max(100).default(20),
 });
 
+/**
+ * Schemas for blogs.
+ */
+export const PriceSchema = z.object({
+  id: z.number(),
+  price: z.string(),
```
Alikznollet:
```suggestion
  price: z.float32(),
```

Alikznollet (changes requested):

Alikznollet (approved):
All good now!

PR #199, opened 2026-03-10
Server documentation

MrLegitFTW:
## 🎯 MAIN GOAL
* adds documentation on how to deploy to the server + a layout 

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

**Missing documentation:**
* Will update README with an installation guide and some more main information

## 🔍 HOW TO TEST
1. go to github pages and :eye: 

## ⚠️ REMAINING ISSUES

## ✅ CLOSED ISSUES
- #197


docs/server/database.md
```diff
@@ -0,0 +1,112 @@
...
+For an interactive overview, open `server/database/db.dbml` with a DBML viewer (e.g.
+the [DBML viewer plugin](https://marketplace.visualstudio.com/items?itemName=matt-meyers.vscode-dbml) for VS Code).
+
+![Database svg](database.svg)
+
+### Core Tables
+
+**`productions`** – The central table. Represents a show or performance, with multilingual JSONB fields for title,
+subtitle, descriptions, artist, tagline, and credits. Also holds `performer_type`, `attendance_mode`, and a
+`planning_id` for external planning system references.
```
landgeet: I think we should remove planning ID here

docs/server/database.svg
landgeet, line 1: Some primary keys are highlighted as not null, others aren't.
Maybe we should make all of them non-nullable (even though pk implies this)?

landgeet (changes requested):
Looks really good, some general remarks:
- the images look really small and barely readable, so maybe a different them would help here, you can try some and send me how it looks.
- maybe you should remove the planningID from the database scheme because we don't use this anymore
- Looking at the database scheme, i realised we made some primary keys NOT NULL and the other we didn't, we should be consistent here (even though pk implicates not null)

docs/server/database.md
MrLegitFTW: We should indeed

Alikznollet (commented):
Vergeet niet nog eens dev te mergen.

Alikznollet (approved):

landgeet (approved):

PR #202, opened 2026-03-11
Fix Scraper + Logger

Alikznollet:
## 🎯 MAIN GOAL

* The scraper had some weird stuff happening with it that should now be fixed, weird stuff includes:
  * Crashing when invalid dates were scraped from the VNV api (why were they there in the first place??)
  * Crashing because I had mistyped a URL...

* The logger needed a path to be defined so it knows where to put logs. That is now fixed by adding a variable `LOGGER_PATH` to `.env` with the **absolute** path to the place you want the logs to be saved.

Logs for the backend will be saved as `backend-{datetime}.log`.

## ✅ CLOSED ISSUES

No issues had been made for this.


MrLegitFTW (approved):
This should fix it. Will be able to tell when we roll-out on prodution.

Issue #171, opened 2026-03-07
[HEADER] general header


ddndauw:
- [ ] Create a header component.

PR #229, opened 2026-03-14
Header

jillvanham:
## 🎯 MAIN GOAL
- [ ] this PR implements the general header

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* all.

## 📝 DOCUMENTATION

**Missing documentation:**

## 🔍 HOW TO TEST
1. Type npm i in the root.
2. Start the localhost with npm run start:dev.
3. Check the header. Resize your browser window. Verify that the header elements adapt correctly and the hamburger menu appears on smaller screens.
4. Check the admin header by manually setting the admin state to true in Header.vue (const isAdmin = ref(true)).
5. Check out the light-dark mode and the language button.

## ⚠️ REMAINING ISSUES
* explain where the bugs are located
* clarify what needs to be changed

## ✅ CLOSED ISSUES
- #171


Alikznollet:
The header already looks very good! The different versions for smaller screens are very nice too. I didn't look at the code since you mentioned you still had to clean it up.

ddndauw:
Visual wise the header looks really good! The transition to the hamburger menu for smaller screen sizes goes very smoothly!

ddndauw:
line

Since the admin version has no tabs in the header, the line on the smaller screen sizes looks a bit off, I do not really know why. Maybe because the buttons are really "visible" now compared to the logo? When there are tabs, the focus is on the tabs, but here it is a bit too much on the buttons? I do like how it looks on the non-admin version though !

frontend/app/components/AppHeader.vue
```diff
@@ -0,0 +1,221 @@
...
+  lastScrollPosition.value = currentScroll
+}
+
+onMounted(() => {
+  window.addEventListener('scroll', handleScroll)
+  // Sync theme state with DOM on mount
+  if (document.documentElement.classList.contains('dark')) {
+    isDark.value = true
+  }
+})
```
Alikznollet, line 135: One thing I noticed is that when you scroll down the header dissapears and appears correctly if you scroll back up. But say when you click on a production and then press the `back` button on your browser the header will **always** dissapear when you get back to the page. Maybe it's a best practise to always show the header when you get back to the page even though you're in the middle of it? I know there's also a  "pageshow" eventlistener so maybe that could be of use?

frontend/app/components/AppHeader.vue
```diff
@@ -0,0 +1,221 @@
...
+  font-weight: 900;
+  cursor: pointer;
+  outline: none;
+  transition: all 0.2s ease;
+}
+
+.btn-outline:hover {
+  background: var(--foreground);
+  color: var(--background);
+}
```
Alikznollet: These could maybe be re-used when put in the main css file? Unless we're not planning on using the same styled buttons in other places.

Alikznollet (changes requested):
Looks very good! I've found one issue that could be seen as a `bug` but I'm not quite sure myself, and something that could be made re-usable. 

Visually it's all really good!

Alikznollet (approved):
Looks real good! Just a weird merge conflict that is probably just github being weird you'll have to fix but after that definitely ready to merge!

Robindr145 (approved):
this looks really solid, no weird bugs found !

Issue #236, opened 2026-03-17
Update workflows


MrLegitFTW:
- split the workflows into multiple parts

PR #237, opened 2026-03-17
Worker update

MrLegitFTW:
## 🎯 MAIN GOAL
* updates the workflows

## 🛠️ TESTING
- [ ] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. this pr

## ⚠️ REMAINING ISSUES

## ✅ CLOSED ISSUES
- #236


.github/workflows/deploy-preview.yml
```diff
@@ -1,19 +1,22 @@
...
-  workflow_dispatch:
+  workflow_run:
+    workflows: [ "Backend Tests", "Frontend Tests" ]
+    types:
+      - completed
 
 jobs:
-  build-and-deploy:
+  deploy:
+    if: ${{ github.event.workflow_run.conclusion == 'success' }}
```
Alikznollet: If I'm understanding correctly the preview only runs when all tests succeed?

Do we want this? Because technically it could be that the tests are cooked because of a change but the preview can still run.

Alikznollet (commented):
Just a small comment about running the workers.

.github/workflows/deploy-preview.yml
MrLegitFTW: I am testing a couple of configs to see what works best. The one I have landed on will just deploy regardless of the tests outcome.

Alikznollet (approved):
Looks good to me. Good split of workers.

PR #238, opened 2026-03-17
Turn Paginated Objects Generic

### LINKED ISSUE #221: Second look at PaginatedResponse
This is a low priority issue!

- [ ] Have a look at the PaginatedResponse and try to make it generic.

Alikznollet:
## 🎯 MAIN GOAL
- [x] this PR refactors part of the backend

#### Changelog

* Refactored all needed things to now use PaginatedResponse<T> for Paginated responses. (this includes both backend frontend and common)
* Fixed the linter in `frontend` and `frontend/tests`.

## 🛠️ TESTING
- [x] all tests succeed

**Notes:**
* Nothing in tests should have changed.

## 📝 DOCUMENTATION

**Missing documentation:**
* Still have to look into what docs to change.

## 🔍 HOW TO TEST

Open Swagger and have a look inside things that return a Paginated object, look at the schemas.

## 📸 SCREENSHOTS:

afbeelding

## ✅ CLOSED ISSUES
- Closes #221


common/src/objects/pagination.ts
landgeet, line 1: I would consider maybe renaming this file to something like `pagination.ts` because `extra.ts` seems a bit to generic.

landgeet (approved):
Looks good to me, just some remarks/questions:
- This might be a dumb question, but why do we need to wrap in this `PaginatedResponse` exactly?
- Maybe we should make a little segment in the docs explaining the pagination?

common/src/objects/pagination.ts
Alikznollet: Will do!

Alikznollet:
>  * This might be a dumb question, but why do we need to wrap in this `PaginatedResponse` exactly?

When we want to retrieve a list that contains *a lot* of elements (like all productions for example), we don't want to send the whole list in one single package to the user. This would take way too long and would take way too much bandwidth too.

The solution is to paginate the responses but then the user also needs to know some metadata like:
* `page`
* `limit`
* `totalItems`

This way the user can know what page they are on and how many pages (and therefore items) are still left.

>  * Maybe we should make a little segment in the docs explaining the pagination?

This is a very good suggestion, I'll look into doing that this afternoon.

landgeet:
Alright makes sense, thanks!

Alikznollet:
> Alright makes sense, thanks!

@landgeet I wrote documentation for the pagination. If you have any time could you check it out?

landgeet (changes requested):
Make sure to link the  `paginated` text in the api reference doc to the pagination doc

landgeet (approved):
looks good now, ready to merge!

MatthieuDC (approved):
Everything looks correct! 

The only thing I don't get is why e.g. tags from productions don't need the same PaginatedResponse wrapper? Is it because the amount of objects attached like this doesn't get very big?

Alikznollet:
> Everything looks correct!
> 
> The only thing I don't get is why e.g. tags from productions don't need the same PaginatedResponse wrapper? Is it because the amount of objects attached like this doesn't get very big?

Exactly this, anything regarding tags, blogs, prices and location all usually won't have more than 20 items (unless their data is fucked). Having a pagination wrapper here would thus be a little overkill.

PR #240, opened 2026-03-18
Feature/reusable parser + templates

### LINKED ISSUE #231: finish CSV parser and create csv templates
- finish parser with pagination and adding necesarry functions
- add csv templates for uses

landgeet:
## 🎯 MAIN GOAL
* create templates for csv parsing
* make parser functions to easily parse and inject csvs with new data

Some disclaimers/design choices for templates:
- Events and productions need an id to link the locations, tags, prices and blogs to them. These events/productions will have autogenerated ids in the database and the old ID will be placed in their `legacy_id` as `"csv-{ID}"`
- Prices have a seperate CSV file because they have multiple fields with translations and would be really messy to put it in events.
- Tags have a seperate CSV file for these reasons:
  - This opens up the ability to add batches of tags and easily link them to mulitple productions.
  - It adds the posibility to add tags after inserting productions instead of needing to instantly couple a tag to the productions.
  - Adding them to the productions csv would create a mess and would add more room for human error with translating mutliple names and having them in the same order in a list.
> [!IMPORTANT]
> For this seperation of prices/tags to work, we need to makes sure that they only use the ids used in previous CSV-files (a.k.a. the legacy_ids with `csv-` in front). Otherwise this would create a huge mess.

If you have any questions surrounding these decisions, feel free to reach out!

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

**Missing documentation:**
* work in progress

## 🔍 HOW TO TEST
- templates: look at the templates and looks if everything makes sense and is correct (and if nothing is missing)
- parser:
  - Run the tests and see if they succeed
  - Look at the testing csv files i added and add/change more data to test more
  - If you want to test the inserting function, you can do the following: 
```bash
  npm run scraper:reset  # clear the dev db
  npm run scraper:inject:structured:test   # inject the testing data
```
- Now run some queries on the database to check if the data was inserted correctly.
- Don't forget to clear the database after and run the scraper again.

## ⚠️ REMAINING ISSUES
* Translation needs to be added still (see TODO in code)
* Put this in a service.

## ✅ CLOSED ISSUES
- Closes #231


landgeet:
As always with random test data and run scripts, i can remove them just before merging.

MatthieuDC (approved):
This looks very good. Nice the old csv parser is still there and used in some cases. The templates are also very nice to have. Tests all look good and working as well

MrLegitFTW (commented):
If we want to use translate here, then we would need to put the parser into a nest-module. I would guess we want to still do this anyways seeing as it is good practice to do so? Are we planning on putting this into a module?

landgeet:
> If we want to use translate here, then we would need to put the parser into a nest-module. I would guess we want to still do this anyways seeing as it is good practice to do so? Are we planning on putting this into a module?

yes, i will do this in a seperate pr.

MrLegitFTW (approved):
This looks good to me, I am confident this works well, and like the old-new dynamic you have going on here. Only worried you might have to switch to using modules seeing as on dev the util module has been expanded upon and is now slightly different. I am fine with it being fixed like this now when you merge from dev and doing that in a seperate pr. I know you want to get this parser behind you but am not sure you are free just yet :-(

Figure out how you want to implement that but either way is fine by me.

Issue #223, opened 2026-03-13
Database Service for Galleries, GalleryItems and Crops.


Alikznollet:
- [ ] Create the DatabaseService for these items.

Issue #222, opened 2026-03-13
Database Update for Galleries, GelleryItems and Crops.


Alikznollet:
- [ ] This is the way VNV saves their objects for media.
- [ ] We need to simplify their objects into the bare things we need.

PR #246, opened 2026-03-21
Media db

MrLegitFTW:
## 🎯 MAIN GOAL

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. for documentation the db docs such as backupscript, dbml and githu docs should all be up-to-date
2. for the db itself testing is not possible until the endpoints are made. you can always check via psql.

## ⚠️ REMAINING ISSUES

## ✅ CLOSED ISSUES
- #222 
- #223

### DIFF
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -1,6 +1,6 @@
 import { BadRequestException, Injectable } from "@nestjs/common";
 import { DbService } from "./db.service";
-import { BlogDto, CreateBlogDto, UpdateBlogDto } from "../dto/dto";
+import { BlogDto, CreateBlogDto, MediaGalleryDto, UpdateBlogDto, } from "../dto/dto";
 import { ResourceGoneException } from "../common/exceptions";
 import { PaginatedResponse } from "@repo/common";
 
@@ -186,5 +186,20 @@ export class BlogDatabaseService {
     }
   }
 
-  // insert extra functions here if desired.
+  /**
+   * Gets all media galleries linked to a given blog.
+   * @param blog_id The ID of the blog you want.
+   * @returns List of MediaGalleryDto linked to the blog.
+   */
+  async getMediaFromBlog(blog_id: number): Promise<MediaGalleryDto[]> {
+    const query = `
+      SELECT mg.id, mg.legacy_id, mg.name, mg.created_at, mg.updated_at
+      FROM media_gallery mg
+      INNER JOIN blog_media_gallery bmg ON bmg.gallery_id = mg.id
+      WHERE bmg.blog_id = $1
+      ORDER BY mg.id
+    `;
+
+    return this.db.query<MediaGalleryDto>(query, [blog_id]);
+  }
 }
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -188,7 +188,7 @@ export class BlogDatabaseService {
   }
 
   /**
-   * Gets all media galleries linked to a given blog.
+   * Gets media gallery linked to a given blog.
    * @param blog_id The ID of the blog you want.
    * @returns List of MediaGalleryDto linked to the blog.
    */
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -27,6 +27,7 @@ export class BlogDatabaseService {
     if (result.length === 0) {
       throw new ResourceGoneException(`Blog with ID ${id} not found`);
     }
+
     return result[0];
   }
 
@@ -191,7 +192,7 @@ export class BlogDatabaseService {
    * @param blog_id The ID of the blog you want.
    * @returns List of MediaGalleryDto linked to the blog.
    */
-  async getMediaFromBlog(blog_id: number): Promise<MediaGalleryDto[]> {
+  async getMediaFromBlog(blog_id: number): Promise<MediaGalleryDto> {
     const query = `
       SELECT mg.id, mg.legacy_id, mg.name, mg.created_at, mg.updated_at
       FROM media_gallery mg
@@ -200,6 +201,47 @@ export class BlogDatabaseService {
       ORDER BY mg.id
     `;
 
-    return this.db.query<MediaGalleryDto>(query, [blog_id]);
+    const result = await this.db.query<MediaGalleryDto>(query, [blog_id]);
+
+    if (result.length === 0) {
+      throw new ResourceGoneException(
+        `Blog with ID ${blog_id} has no media gallery`,
+      );
+    }
+
+    return result[0];
+  }
+
+  /**
+   * Links a media gallery to a given blog.
+   * @param blog_id The ID of the blog to link to.
+   * @param gallery_id The ID of the media gallery to link.
+   * @returns void
+   */
+  async linkMediaToBlog(blog_id: number, gallery_id: number): Promise<void> {
+    const query = `
+    INSERT INTO blog_media_gallery (blog_id, gallery_id)
+    VALUES ($1, $2)
+    ON CONFLICT DO NOTHING
+  `;
+
+    await this.db.query(query, [blog_id, gallery_id]);
+  }
+
+  /**
+   * Unlinks a media gallery from a given blog.
+   * @param blog_id The ID of the blog.
+   * @param gallery_id The ID of the media gallery to unlink.
+   */
+  async unlinkMediaFromBlog(
+    blog_id: number,
+    gallery_id: number,
+  ): Promise<void> {
+    const query = `
+    DELETE FROM blog_media_gallery
+    WHERE blog_id = $1 AND gallery_id = $2
+  `;
+
+    await this.db.query(query, [blog_id, gallery_id]);
   }
 }
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -194,7 +194,7 @@ export class BlogDatabaseService {
    */
   async getMediaFromBlog(blog_id: number): Promise<MediaGalleryDto> {
     const query = `
-      SELECT mg.id, mg.legacy_id, mg.name, mg.created_at, mg.updated_at
+      SELECT mg.id, mg.name, mg.created_at, mg.updated_at
       FROM media_gallery mg
       INNER JOIN blog_media_gallery bmg ON bmg.gallery_id = mg.id
       WHERE bmg.blog_id = $1
--- a/backend/src/database/db.media.service.ts
+++ b/backend/src/database/db.media.service.ts
@@ -340,6 +340,35 @@ export class MediaDatabaseService {
     }
   }
 
+  /**
+   * Link an existing item to an existing gallery.
+   * Silently does nothing if the link already exists (ON CONFLICT DO NOTHING).
+   * @param galleryId The gallery to link to.
+   * @param itemId The item to link.
+   */
+  async linkItemToGallery(galleryId: number, itemId: number): Promise<void> {
+    await this.db.query(
+      `INSERT INTO gallery_item (gallery_id, item_id) VALUES ($1, $2) ON CONFLICT DO NOTHING`,
+      [galleryId, itemId],
+    );
+  }
+
+  /**
+   * Unlink an item from a gallery. Does not delete the item itself.
+   * Silently does nothing if the link doesn't exist.
+   * @param galleryId The gallery to unlink from.
+   * @param itemId The item to unlink.
+   */
+  async unlinkItemFromGallery(
+    galleryId: number,
+    itemId: number,
+  ): Promise<void> {
+    await this.db.query(
+      `DELETE FROM gallery_item WHERE gallery_id = $1 AND item_id = $2`,
+      [galleryId, itemId],
+    );
+  }
+
   // ----------------------------------------------------------------
   // Media Crop
   // ----------------------------------------------------------------
--- a/backend/src/database/db.media.service.ts
+++ b/backend/src/database/db.media.service.ts
@@ -1,0 +1,485 @@
+import { BadRequestException, Injectable } from "@nestjs/common";
+import { DbService } from "./db.service";
+import {
+  CreateMediaCropDto,
+  CreateMediaGalleryDto,
+  CreateMediaItemDto,
+  MediaCropDto,
+  MediaGalleryDto,
+  MediaItemDto,
+  UpdateMediaCropDto,
+  UpdateMediaGalleryDto,
+  UpdateMediaItemDto,
+} from "../dto/dto";
+import { ResourceGoneException } from "../common/exceptions";
+import { PaginatedResponse } from "@repo/common";
+
+@Injectable()
+export class EventMediaService {
+  // need to give a db service as param when used. -> see db.service.
+  constructor(private db: DbService) {}
+
+  // ----------------------------------------------------------------
+  // Media Gallery
+  // ----------------------------------------------------------------
+
+  /**
+   * Get a single gallery by its ID.
+   * @param id The ID we're trying to fetch.
+   * @returns The MediaGallery if there is one.
+   */
+  async getGalleryById(id: number): Promise<MediaGalleryDto> {
+    const query = `
+      SELECT id, legacy_id, name, created_at, updated_at
+      FROM media_gallery
+      WHERE id = $1
+    `;
+
+    const result = await this.db.query<MediaGalleryDto>(query, [id]);
+
+    return result[0];
+  }
+
+  /**
+   * Get galleries with pagination.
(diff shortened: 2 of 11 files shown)


backend/src/database/db.blog.service.ts
```diff
@@ -186,5 +186,20 @@ export class BlogDatabaseService {
...
+    const query = `
+      SELECT mg.id, mg.legacy_id, mg.name, mg.created_at, mg.updated_at
+      FROM media_gallery mg
+      INNER JOIN blog_media_gallery bmg ON bmg.gallery_id = mg.id
+      WHERE bmg.blog_id = $1
+      ORDER BY mg.id
+    `;
+
+    return this.db.query<MediaGalleryDto>(query, [blog_id]);
+  }
```
Alikznollet: Are we planning on being able to link multiple galleries to one blog? Because that would be more like just linking multiple items to a gallery right? Maybe this could be more clear if it just returned one single gallery just like with Locations.

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,503 @@
...
+  MediaItemDto,
+  UpdateMediaCropDto,
+  UpdateMediaGalleryDto,
+  UpdateMediaItemDto,
+} from "../dto/dto";
+import { ResourceGoneException } from "../common/exceptions";
+import { PaginatedResponse } from "@repo/common";
+
+@Injectable()
+export class EventMediaService {
```
Alikznollet:
```suggestion
export class MediaDatabaseService {
```

I think you forgot to alter the name.

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,503 @@
...
+  /**
+   * Create a new media item and link it to a gallery.
+   * @param item Must be of type CreateMediaItemDto.
+   * @param galleryId The gallery to link this item to.
+   * @returns The created media item.
+   */
+  async createItem(
+    item: CreateMediaItemDto,
+    galleryId: number,
+  ): Promise<MediaItemDto> {
```
Alikznollet, line 186: I don't know if having the creation and linking of a `MediaItem` be in the same function is a good thing. What if we want to link an item to more than one gallery?

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,503 @@
...
+  /**
+   * Create a new crop and link it to a media item.
+   * @param crop Must be of type CreateMediaCropDto.
+   * @param itemId The item to link this crop to.
+   * @returns The created crop.
+   */
+  async createCrop(
+    crop: CreateMediaCropDto,
+    itemId: number,
+  ): Promise<MediaCropDto> {
```
Alikznollet, line 405: So I see below here that there is actually manual linking for crops, so I'd maybe suggest removing it from the create, this way we can seperate these to have more control over what we do.

backend/src/database/db.production.service.ts
Alikznollet, line 1: Are we missing functions for linking a gallery to a production?

backend/src/database/db.production.service.ts
```diff
@@ -586,4 +587,21 @@ export class ProductionDatabaseService {
...
+    const query = `
+      SELECT mg.id, mg.legacy_id, mg.name, mg.created_at, mg.updated_at
+      FROM media_gallery mg
+      INNER JOIN production_media_gallery pmg ON pmg.gallery_id = mg.id
+      WHERE pmg.production_id = $1
+      ORDER BY mg.id
+    `;
+
+    return this.db.query<MediaGalleryDto>(query, [prod_id]);
+  }
```
Alikznollet, line 649: Here again like for blogs maybe consider changing this to one single gallery returned, because galleries are already a collection of items.

backend/src/database/db.blog.service.ts
Alikznollet, line 1: Forgot to add the linking functions?

common/src/objects/media.ts
```diff
@@ -0,0 +1,73 @@
+import z from "zod";
+
+// Gallery
+export const MediaGallerySchema = z.object({
+  id: z.number(),
+  name: z.string(), // TODO this needs language?
+  created_at: z.iso.datetime(),
+  updated_at: z.iso.datetime(),
+  legacy_id: z.string(),
```
Alikznollet: Like with all other API schemas remove legacy_id here so it is not shown/editable via the API. Don't forget to change the DB function too then.

common/src/objects/media.ts
```diff
@@ -0,0 +1,73 @@
...
+  id: z.number(),
+  type: z.string(),
+  original_filename: z.string(),
+  position: z.number(),
+  width: z.number(),
+  height: z.number(),
+  format: z.string(),
+  created_at: z.iso.datetime(),
+  updated_at: z.iso.datetime(),
+  legacy_id: z.string(),
```
Alikznollet: Same as above comment.

common/src/objects/media.ts
```diff
@@ -0,0 +1,73 @@
...
+export type UpdateMediaItem = z.infer<typeof UpdatedMediaItemSchema>;
+
+// crop
+export const MediaCropSchema = z.object({
+  id: z.number(),
+  name: z.string(), // TODO this needs language?
+  url: z.string(),
+  created_at: z.iso.datetime(),
+  updated_at: z.iso.datetime(),
+  legacy_id: z.string(),
```
Alikznollet: Same as above comment.

Alikznollet (changes requested):
The code and database schemas look great overall. Some forgotten code or small changes I've requested and some things we might still have to discuss.

backend/src/database/db.blog.service.ts
MrLegitFTW: Yeah wasn't sure on that, will update.

backend/src/database/db.media.service.ts
MrLegitFTW: Jup :t-rex:

common/src/objects/media.ts
Alikznollet: Also to follow up on the highlighted code here, maybe it does need language support if we want to show the name somewhere, if not this could just stay an internal thing.

MrLegitFTW: Will do, forgot we changed that.

backend/src/database/db.blog.service.ts
MrLegitFTW: No should be at the bottom?

MrLegitFTW: ah no they are in the media db service, I am guessing we would prefer them here instead?

backend/src/database/db.production.service.ts
MrLegitFTW: same as blog.

backend/src/database/db.production.service.ts
MrLegitFTW: :taco:

backend/src/database/db.blog.service.ts
Alikznollet: Yeah I think linking for `production` + `gallery` and `blog` + `gallery` should be in their respective services to keep it consistent with the other endpoints and services.

backend/src/database/db.media.service.ts
MrLegitFTW: then we create it first and link it to another via the seperate linking function?

MrLegitFTW: Would we prefer 1 create function where you can link to multiple galleries instead or one where we do not link at all then? I would suggest we allow the linking to a list then?

Alikznollet: I would say we just have a function that lets you create an item and one that lets you link an item to a gallery.

backend/src/database/db.media.service.ts
MrLegitFTW: That is an option however creating an item without linking it could lead to bugs where it could be easy to forget to link it? therefore I suggest a list to solve the problem of linking to multiple  galleries.

backend/src/database/db.media.service.ts
Alikznollet: And if you want to link an item to multiple galleries just call that function multiple times with different gallery ids. Though now that I think about it I can't think of a use off the top of my head.

backend/src/database/db.media.service.ts
Alikznollet: You know what, that's actually a great idea! You can do that!

MrLegitFTW: will do

backend/src/database/db.blog.service.ts
```diff
@@ -186,5 +187,61 @@ export class BlogDatabaseService {
     }
   }
 
-  // insert extra functions here if desired.
+  /**
+   * Gets all media galleries linked to a given blog.
```
Alikznollet: Rather nitpicky I know but could you alter the comment too?

backend/src/database/db.media.service.ts
Alikznollet, line 1: Are there manual linking functions for galleries and items? Also I'm not sure if we'll have to link multiple crops to different items because each item is supposed to be one image. The list for galleries is good though.

backend/src/database/db.production.service.ts
```diff
@@ -586,4 +587,64 @@ export class ProductionDatabaseService {
       );
     }
   }
+
+  /**
+   * Gets all media galleries linked to a given production.
```
Alikznollet: Same as before, could you change the comment again?

Alikznollet (changes requested):
Some more comments. Could you let me know when it's safe for me to start work on endpoints?

backend/src/database/db.media.service.ts
MrLegitFTW: I'll remove the list create for crops then, does make sense. Yes there are linking functions for linking galleries, items and crops individually.

Alikznollet:
@MrLegitFTW While working on the endpoints I noticed you forgot to register the new DBService in the module.

Alikznollet:
@MrLegitFTW Also cannot find any seperate functions for linking items to galleries, only the item creation function in the database service.

backend/src/database/db.blog.service.ts
MrLegitFTW: Don't quite understand what you want here?

backend/src/database/db.production.service.ts
MrLegitFTW: Same here, do not quite understand what it is you want?

backend/src/database/db.blog.service.ts
Alikznollet: It says gets ALL media galleries from ...

And the function now only returns one

MrLegitFTW:
please note that the name fields have been updated to JSONB in the db so please pay attention to that when making the endpoints

backend/src/database/db.blog.service.ts
MrLegitFTW: ah yes very true, will change that

Alikznollet (approved):
This looks good to me!

Alikznollet (changes requested):
Wait nevermind my previous comment, we're missing functions to get all items and all crops. These are standard REST things so we should include them.

Alikznollet:
> please note that the name fields have been updated to JSONB in the db so please pay attention to that when making the endpoints

Maybe also time for me to adress this, their API does not define multiple languages for these name fields so I'm not sure if we want this. Maybe need some other people's opinions on this.

Alikznollet:
> Wait nevermind my previous comment, we're missing functions to get all items and all crops. These are standard REST things so we should include them.

Also, for crops this might be overkill and we don't need this. Only for items, since then in the frontend we could make an overview of all items in the CMS and users could link items to a certain gallery.

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   */
+  async deleteGallery(id: number): Promise<void> {
+    const result = await this.db.query(
+      `DELETE FROM media_gallery WHERE id = $1 RETURNING id`,
+      [id],
+    );
+    if (result.length == 0) {
+      throw new ResourceGoneException(`Cannot delete: Gallery ${id} not found`);
+    }
+  }
```
landgeet, line 120: I know this is really nitpicky, but it would be nicer if we had the following structure everywhere:

```suggestion
  async deleteGallery(id: number): Promise<void> {
    const query = `DELETE FROM media_gallery WHERE id = $1 RETURNING id`;
    const result = await this.db.query(query, [id]);
    if (result.length == 0) {
      throw new ResourceGoneException(`Cannot delete: Gallery ${id} not found`);
    }
  }
```
So basically have the query as seperate variable and the result as seperate variable.
This is just for consistency reasons :)

Maybe we should make an issue for looking through every file to see if everything is consistent between the files? (this could be for milestone 3 when we do cleanup)

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   */
+  async deleteItem(id: number): Promise<void> {
+    const result = await this.db.query(
+      `DELETE FROM media_item WHERE id = $1 RETURNING id`,
+      [id],
+    );
+    if (result.length == 0) {
+      throw new ResourceGoneException(`Cannot delete: Item ${id} not found`);
+    }
+  }
```
landgeet, line 313: Here too

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   * Silently does nothing if the link already exists (ON CONFLICT DO NOTHING).
+   * @param galleryId The gallery to link to.
+   * @param itemId The item to link.
+   */
+  async linkItemToGallery(galleryId: number, itemId: number): Promise<void> {
+    await this.db.query(
+      `INSERT INTO gallery_item (gallery_id, item_id) VALUES ($1, $2) ON CONFLICT DO NOTHING`,
+      [galleryId, itemId],
+    );
+  }
```
landgeet, line 324: same here

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   */
+  async unlinkItemFromGallery(
+    galleryId: number,
+    itemId: number,
+  ): Promise<void> {
+    await this.db.query(
+      `DELETE FROM gallery_item WHERE gallery_id = $1 AND item_id = $2`,
+      [galleryId, itemId],
+    );
+  }
```
landgeet, line 338: same here

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   */
+  async deleteCrop(id: number): Promise<void> {
+    const result = await this.db.query(
+      `DELETE FROM media_crop WHERE id = $1 RETURNING id`,
+      [id],
+    );
+    if (result.length == 0) {
+      throw new ResourceGoneException(`Cannot delete: Crop ${id} not found`);
+    }
+  }
```
landgeet, line 493: same here

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   * Silently does nothing if the link already exists (ON CONFLICT DO NOTHING).
+   * @param itemId The item to link to.
+   * @param cropId The crop to link.
+   */
+  async linkCropToItem(itemId: number, cropId: number): Promise<void> {
+    await this.db.query(
+      `INSERT INTO item_crop (item_id, crop_id) VALUES ($1, $2) ON CONFLICT DO NOTHING`,
+      [itemId, cropId],
+    );
+  }
```
landgeet, line 504: same here

backend/src/database/db.media.service.ts
```diff
@@ -0,0 +1,543 @@
...
+   * Silently does nothing if the link doesn't exist.
+   * @param itemId The item to unlink from.
+   * @param cropId The crop to unlink.
+   */
+  async unlinkCropFromItem(itemId: number, cropId: number): Promise<void> {
+    await this.db.query(
+      `DELETE FROM item_crop WHERE item_id = $1 AND crop_id = $2`,
+      [itemId, cropId],
+    );
+  }
```
landgeet, line 515: same here

landgeet (changes requested):
This looks really good!
I like that you chose to keep everything in one service so it's bundled in one place.
I have a small nitpicky remark below to keep consistency :-)
If this and the remarks of Alex are fixed, this can be merged!

backend/src/database/db.media.service.ts
MrLegitFTW: yes that seems like a good idea, could you add that to the topics for the next meeting?

MrLegitFTW:
should be updated now, please re-review when you can.

landgeet (approved):
Approved!

Alikznollet (approved):
d

Issue #178, opened 2026-03-07
[STORIES] general story display component


ddndauw:
- [ ] Create a timeline (listview) of stories. Story listview components are defined in #167

Issue #167, opened 2026-03-07
[LIST VIEWS] general story list view component


ddndauw:
- [ ] Create a general story list view component. (That can be used for list views displaying stories later on.)

PR #248, opened 2026-03-22
Feature/general stories

Robindr145:
## 🎯 MAIN GOAL
* added general components for the blogs-page: timeline, list view and single blogs view

>[!WARNING]
> test are not yet implemented
> I'm not a huge fan of the darkmode atm, but this will be the main goal for another issue
## 🛠️ TESTING
- will be solved in another PR

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
- `npm run start:dev` and go to `baseURL/stories` or navigate to the stories-section via the homepage
- check out the light/dark -mode version as well as the language
- sort the blogs, click on a blog, search for a blog

## ⚠️ REMAINING ISSUES
- dark mode needs some improvements

## ✅ CLOSED ISSUES
- #167 
- #178


jillvanham:
Everything looks really good! I love the feature where you can expand and collapse the stories for a specific year. I also noticed you've already accounted for stories without images, so that's handled perfectly. Great to see that both the list view and the detail page are implemented and that the stories are already clickable. The detail page looks really great as well.

The only visual remark I have is that everything feels a bit 'boxy' right now. We could solve this with a consistent border-radius, but it's fine to leave that for a separate issue (if we wanted to change that).

Nowak-Dominik:
@Robindr145 
I just reviewed your PR visually and I have to say, it looks sick! I’m a big fan.

However, there are a few small things I’d like to point out.

- Search bar:
image

I think you should follow the layout from the mockup more closely. By that I mean the search bar and filter button should fill the entire width, specifically the same width as the content underneath.

So basically, this whole section:

image

Also, something else. I don’t think this is for you to fix, but I’ll mention it anyway. The filter button currently opens the filter section next to it instead of underneath it. Also, right now there are only "order by" filters, but I think there were more in the mockup if I remember correctly.

Again, this is probably something to be handled in the PR that implements the filters component.

- Empty space

image

The second thing I noticed when I opened your page is that there’s quite a lot of empty space between the search section and the blogs section. Maybe you could reduce it a bit.

It’s not excessively large or anything, but it is noticeable. I think making it slightly smaller would improve the overall feel.

- Timeline

image

I’m a big fan of the timeline, but I think it’s currently too small. It does not take up the full height.

I think it would look better if it were longer, meaning it takes up most or all of the available height.

- Year text

image

image

For some reason, the year label has a hover effect, which makes it feel clickable. But when I click it, nothing happens.

This gives the impression that something is broken. I think it would be better to remove the hover effect since the label isn’t clickable.

- Story view go back button

image

Honestly, when I opened this view, I wanted to go back and it took me a moment to find the button. It’s not visible enough. It looks more like a label.

For example, if you compare the `23 March 2026` label with the `<- back to stories` button, there is not much visual difference.

Also, I think in the mockup the button was placed in the header section, under the title and on the same level as the `23 March 2026    1 min read` labels.

I’d suggest:

* Adding a new back button in that upper section
* Keeping the existing one at the bottom

This way, users can go back quickly from the top and also after finishing the blog, which might be long. That avoids forcing them to scroll all the way up.

Also, making the buttons more visually distinct would help users immediately understand how to navigate back.

- Story view dark mode

image

image

I know dark mode is something that will be improved later overall, but I noticed that the blog body and the section underneath it use different background colors.

In light mode, this is not the case, so the page feels like one cohesive whole. In dark mode, however, the sections feel visually separated.

I think using the same background color here would make the page feel more unified.

Robindr145:
@Nowak-Dominik 
- I'll improve the search-bar length.
- Personally I didn't notice the empty space, so it didn't bother me.
- I'll make the timeline bigger. I made it smaller because it was a bit overwhelming in the beginning, but I'll change it a bit. 
- The year-text should collapse. I think this is a small bug, will fix this.

The individual blogs page: The 2 go back buttons looked a bit weird, but maybe it's because there's not much text (?). I'll improve the overall visibility of the individual blogs page.

The filter button will also include the calendar (which I'm working on), so it'll be less empty. 

Thanks for the suggestions !

jillvanham (commented):
I think everything looks great visually! I really like what you did with the timeline. It would be even better if you extended the timeline all the way to the bottom. I agree with Dominik’s comments, so if you take a look at those, I’m sure it will look even better.

The only thing that could improve is the color scheme. It feels like you’re a little hesitant to use more color contrast. You could compare it to a drawing made only in pencil—everything that's grey (text, lines, etc.) tends to disappear into the background. But again, that's something we can look into in a new issue. Overall, I’m impressed!

ddndauw:
I have checked this PR visually, the timeline looks amazing! I also like how the years, on the left, scroll along, and how the timeline automatically scrolls to the year, when you click one. 

- I also agree with Domi that the search bar should be longer. 
- For the filters, there is a seperate issue for that, somewhere, I think? Because filters aren't only used in stories, but other places as well. I think it would be smart to have a generic filter component, where you can give which filters you want to have/display when using it. Maybe you can claim that issue, if you want it, since you're already busy with it.
- The large space between blogs and search doesn't immediately bother me, but I do not mind it being smaller if the others prefer it that way. 
- I also agree with Domi for the "go back"-button. Also, maybe it's a small detail, but could you put a space after "<-", it sticks a bit too much on the text haha.
go back

- What are these lines for? I've noticed the dots at the beginning of the timeline on the left are connected, but after a while they are not. Is this a bug, or intentional? If it's intentional, why?
dots

I have not noticed any other issues for the moment. Again, great job !!

frontend/app/pages/stories/index.vue
```diff
@@ -0,0 +1,151 @@
...
+    pending.value = true;
+    fetchError.value = null;
+  } else {
+    if (isLoadingMore.value || !hasMore.value) return;
+    isLoadingMore.value = true;
+  }
+
+  try {
+    const raw = await getAll({
+      paginationFilters: { limit: LIMIT, page: page.value, descending: true },
```
Alikznollet: If you want to display NEW -> OLD do `descending: true` and OLD -> NEW `descending: false`. Don't sort this yourself.

Alikznollet: Also you should be asking for `{ lang: [currentLanguage] }` here with the getAll. We designed the backend language feature so the frontend would know what they get.

frontend/app/pages/stories/index.vue
```diff
@@ -0,0 +1,151 @@
...
+// ── Sort / filter / search state ─────────────────────────────────────────────
+const sortOrder   = ref<"newest" | "oldest">("newest");
+const showFilter  = ref(false);
+const searchQuery = ref("");                              // ← nieuw
+
+// ── Pagination state ─────────────────────────────────────────────────────────
+const LIMIT        = 20;
+const page         = ref(0);
+const totalItems   = ref(0);
+const stories      = ref<Array<Blog | BlogView>>([]);
```
Alikznollet: This should be `Array<BlogView>` because when you ask for a specific language from the backend it'll return a view and you don't have to filter for language anymore.

frontend/app/components/blogs/StoryListItem.vue
```diff
@@ -0,0 +1,151 @@
...
+                  gradient placeholder derived from the story id.
+  - No id badge : the #N identifier is intentionally omitted from the card.
+-->
+
+<script lang="ts" setup>
+import type { Blog, BlogView } from "@repo/common";
+import { pickPlaceholderGradient } from "~/utils/constants";
+
+const props = defineProps<{
+  story: Blog | BlogView;
```
Alikznollet: Same here, this will *always* be a `BlogView` if you fetched this correctly with language.

frontend/app/components/blogs/StoryListItem.vue
```diff
@@ -0,0 +1,151 @@
...
+/** Localised date string. */
+const formattedDate = computed(() => {
+  if (!props.story.created_at) return "";
+  const loc = locale.value === "nl" ? "nl-BE" : "en-GB";
+  return new Date(props.story.created_at).toLocaleDateString(loc, {
+    year: "numeric",
+    month: "long",
+    day: "numeric",
+  });
+});
```
Alikznollet: All of these locale things are thus not needed anymore if you fetch with the language provided.

frontend/app/components/blogs/StoryTimeline.vue
```diff
@@ -0,0 +1,144 @@
...
+        : typeof desc === "string"
+          ? desc
+          : (desc as any)?.[locale.value] ?? (desc as any)?.nl ?? (desc as any)?.en ?? "";
+
+    return (
+      title.toLowerCase().includes(q) ||
+      descText.toLowerCase().includes(q)
+    );
+  });
+});
```
Alikznollet: Same thing here, you get the language from the backend, no need to call the specific languages here.

Alikznollet (changes requested):
Quickly skimmed through the code and noticed some stuff that isn't really how it should be.

Firstly try to not filter any things in the frontend. I saw some filtering function but you can't really filter pages of items if you don't have the whole.

Now about the `locale` things. We made sure the backend has functionality to query for `language`. Don't forget about this. The only place where you have to pass the language is in the `getAll` function. After that you are sure you'll get a `BlogView` and that means you can also just call the individual fields of the view safely.

Nowak-Dominik:
@Robindr145 
I just reviewed your PR visually.
Like before, it looks good!

But again I have some remarks.

## **Timelines first dot/point**

It bothers me that the first one looks like it's cut off at the top. Try to adjust it so the dot/point is fully visible.

image

## **Header**
Try to make it bigger and add some description. Right now, you only have the label "Verhalen".
Try to match it more closely to the mockup.

image

Mockup:
image

## **Search section**
Here, you should also add more spacing above and below the search bar. By that, I mean increasing the space between the search bar and the lines above and below it.

image

Another issue is that the "Order by" dropdown is not the same size as the search bar, which makes it look inconsistent.

image

## **Timeline functionality**

I think that when the user clicks on a dot, it should not filter to only that year, but instead scroll to that year.

Later, when you add filter options to show a specific year, this would otherwise duplicate the same functionality. It would be better to use the timeline as a navigation tool rather than a filter.

It may also not be immediately clear to users that they need to click the dot again to see all results. That’s why I think it would be better to make it behave more like a “navigate to” action.

## **Go back to top button**

Honestly, I didn’t even notice it for quite a while, so I initially thought it wasn’t implemented. It’s currently too hidden in the corner.

Maybe try to make it more visible. For example, make it a larger rounded icon button that matches the style of the timeline dots. You could either:

* keep the current style (purple border, no background, purple icon), or
* make it a fully purple button with a white icon

Also, the button is currently fixed to the bottom-right corner on all screen sizes, which makes it even less noticeable on larger screens. Try to keep it within the main content area, centered like we discussed in the meeting today.

image

## **Go back button on blog view page**

You improved the button, and I like it, but maybe there was a misunderstanding about its position. I would like you to place it on the right side, as shown below.

image

Nowak-Dominik:
I think that the current look of the page is for sure good enough and in the future we can adress some small viasual fixes.
Anyway this looks still sick! Byt this I approve this PR visually and I'am going to check the code now. Thank you for the changes you dit.

Alikznollet (approved):
I think code wise this looks good now, seems like you eliminated all of the language problems. Could always have missed something but we can always catch that later.

frontend/app/components/blogs/StorySkeleton.vue
```diff
@@ -0,0 +1,36 @@
+<!--
+  components/blogs/StorySkeleton.vue
+  ====================================
+  Animated loading skeleton voor de eerste page fetch.
```
Nowak-Dominik: Write the comment in English ;)

frontend/app/pages/stories/[id].vue
```diff
@@ -0,0 +1,174 @@
...
+const { title, description: body, image, formattedDate, placeholderGradient: bannerGradient } =
+  useBlogStory(blog);
+
+const readingTime = computed(() => {
+  const words = body.value.split(/\s+/).filter(Boolean).length;
+  return Math.max(1, Math.ceil(words / 200));
+});
+</script>
+
+<template>
```
Comment (Nowak-Dominik) [inline frontend/app/pages/stories/[id].vue:46]: Maybe a suggestion for this template section is that you can split it a little more in separate components becouse it is getting a little big.
I will not request this as a change becouse it is still quite fine but I just wanted to mention it.

frontend/app/utils/formatters.ts
```diff
@@ -1,50 +1,60 @@
 // Formatting helpers used across the frontend (dates and ranges).
 
 // Format ISO date to short string according to locale ('nl' or 'en')
-export function formatDateShort(iso: string, lang: string): string {
+export function formatDateShort(
+  iso: string,
+  lang: string,
+  options: Intl.DateTimeFormatOptions = { day: "numeric", month: "short", year: "numeric" },
```
Nowak-Dominik, line 7: I see you changed this which is fine, but before you merge you should first merge develop to this branch and see if nothing is broken. That is becouse components defined by me (list and grid veiw items for production) use this and after your change they might not use this correctly anymore. And now that the Home page uses my grid view item component, this becomes an issue that might be visible. So please check this out before merging.

Nowak-Dominik (changes requested):
@Robindr145 
Your PR looks great and it's basicly done to be approved, but check quickly the things I mentioned and let me know when it is done so I can check it out quickly and approve this PR. Thanks!

Nowak-Dominik (approved):
The corrections look good to me!
By this I approve this PR. Good job!

PR #251, opened 2026-03-22
Ascending and Descending Queries for Objects

### LINKED ISSUE #244: Asc/Desc filters for Objects
- [ ] Implement Ascending/Descending filter types for Events. (Only for dates).
- [ ] Implement Ascending/Descending filter types for Productions (Only for dates).

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* Added Descending and Ascending order to productions so the user can change these, it is sorted on the **earliest** event `starttime` of that production, all productions with no events are put at **the end** of the list at all times.
* This `descending` boolean field is part of `PaginationFilter` and defaults to `true`.
* Moved pagination filters to their own object in all instances.
* Moved Language queries to their own objects in all instances.
* Reworked the way we call the `getAll` hooks in the frontend so passing optional parameters by name is allowed.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

Might have to update docs for this in some places.

## 🔍 HOW TO TEST

You can test correctly in swagger. Regarding the frontend you can surf to `/temp` and open your console with `F12` to see whether the data is correctly being fetched in pages and such.

## ✅ CLOSED ISSUES
- Closes #244


Alikznollet:
@Nowak-Dominik I slightly reworked the way we call the `getAll()` functions in the Frontend API connection. It was rather hard to pass optional things because you couldn't pass parameters by name, now that is fixed entirely. Could you have a look at this and see if it's fine? I also made a page `/temp` that displays that the fetches still do work and the filters are applied.

Alikznollet:
This is now ready for review!

MrLegitFTW (approved):
This looks good as is, like the consistency we have going on lately and this is another staple of that. You will have to merge from dev first.

backend/src/blog/blog.controller.spec.ts
```diff
@@ -110,7 +111,7 @@ describe("BlogController", () => {
       mockBlogService.getAllBlogs.mockResolvedValue(rawBlogs);
       mockLanguageService.flattenByLanguage.mockReturnValue(flattenedBlogs);
 
-      const result = await controller.getAllBlogs(langQuery, filter);
+      const result = await controller.getAllBlogs(filter, langQuery);
```
landgeet, line 114: Good that you fixed this, but i don't get how all tests could pass before this without it seeing this error. Maybe we should look into this more?

landgeet (approved):
This looks really good! This is a really nice addition to the queries. The code also looks very clean.
There is just one question i have about the fixing of a test.

backend/src/blog/blog.controller.spec.ts
Alikznollet: I also switched the order around in the blogcontroller I believe. That's why this change needed to be made.

landgeet: well i'm stupid, ignore what i said xd

PR #252, opened 2026-03-22
Parser service into module + translations

### LINKED ISSUE #250: Put parser into module
- Put parser into module so the data can be translated.

landgeet:
## 🎯 MAIN GOAL
* make a service for the csv-parser so i can put it in a module.
* do translations for csv data before inserting

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
There is not much to test.
You can try to wipe the DB and run the inject script for the test files and see if they get translated correctly, but this would use our limit of the translator which is not recommended :)

## ⚠️ REMAINING ISSUES
* I still need to create some kind of endpoint do the frontend can acces the parser.

## ✅ CLOSED ISSUES
- Closes #250

### DIFF
--- a/backend/src/csv_parsing/csv_file_parser.spec.ts
+++ b/backend/src/csv_parsing/csv_file_parser.spec.ts
@@ -103,24 +103,6 @@ describe("CSVFileParser", () => {
       });
     });
 
-    it("should handle missing English location with Dutch fallback", () => {
-      const row = {
-        ID: "1",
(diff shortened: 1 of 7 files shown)


Alikznollet (approved):
Looks good, not much to say about this.

MrLegitFTW (approved):
this solves my previous concerns so this is good.

PR #262, opened 2026-03-23
Feature/parser endpoint + composable

### LINKED ISSUE #253: Parser endpoint
- Create endpoint for parser so it can easily be used in the frontend.

landgeet:
## 🎯 MAIN GOAL
* Create endpoints for the parser
  * This required and extra DTO to properly show the body in the Swagger UI
* Change input for parsing functions from filepath to CsvInputSource which can take a Buffer for uploaded files and filepaths as fallback. This makes sure the user can upload csv files and it is handled correctly.
  * This also required the use of file interceptors for the endpoints.
* Implement a composable that can be used in the frontend.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
Check the endpoints in the swagger docs and try to interact with it.

## ⚠️ REMAINING ISSUES
* I need to write documentation about the parser in the MkDocs.

## ✅ CLOSED ISSUES
- Closes #253


backend/src/parser/dto/parser-upload-csv-body.dto.ts
```diff
@@ -0,0 +1,12 @@
+import { ApiProperty, ApiPropertyOptional } from "@nestjs/swagger";
+
```
MrLegitFTW: is there no way we can add this to our normal dto?

MrLegitFTW (commented):
Looks good just a question below.

backend/src/parser/dto/parser-upload-csv-body.dto.ts
landgeet: I can put this in the normal dto file, but i can't use the createZodDto because it has a mutlipart body. Do you want me to add it to there?

MrLegitFTW: I think that would make it easier to find/ more consistent with our current impl.

Alikznollet (approved):
This looks solid to me.

MrLegitFTW (approved):
I am happy with this.

Issue #169, opened 2026-03-07
[GRID VIEWS] general production grid view component


ddndauw:
- [ ] Create a general production grid view component. (That can be used for grid views displaying productions later on.)

PR #266, opened 2026-03-26
Grid view item component for production

Nowak-Dominik:
## 🎯 MAIN GOAL
* Grid view item component for production
* Added some extra fields to the TumbnailPlaceholder
* None

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* Test will be written in different pr's. (Tested by a page that uses this component)

## 📝 DOCUMENTATION

**Missing documentation:**
* None

## 🔍 HOW TO TEST
Run `npm run start:dev` and go to `baseUrl/tempShowcase`
Or
Look at the preview in this pr and go there to `baseUrl/tempShowcase`

## ⚠️ REMAINING ISSUES
* None
* None

## ✅ CLOSED ISSUES
- #169


Alikznollet (approved):
Code for this looks great as does the visuals themselves!

Robindr145 (approved):
This looks really good, but I have just one question, what happens if there are more tags for a production ? Does it show some tags and then a +5-tag e.g. (when there are 5 more tags) ?  Or will it fiill the box with all the tags ?

Nowak-Dominik:
> This looks really good, but I have just one question, what happens if there are more tags for a production ? Does it show some tags and then a +5-tag e.g. (when there are 5 more tags) ? Or will it fiill the box with all the tags ?

It just shows all tags. The container is always big enough to contain them all. The size of the container depends on the items that are inside of it so it will always show all tags and there will be enough space for it. Also we can assume that there never will be a production with like 20+ tags, so the component won't be like way too big or something.

PR #268, opened 2026-03-26
Update -> ModifyX and ReplaceX

### LINKED ISSUE #256: Create Seperate DTOs for Replacing Objects

Alikznollet:
## 🎯 MAIN GOAL
- [x] This PR refactors part of the backend for better code readability and clarity.

### Changelog

* For endpoints that contain a REPLACE endpoint a new object and schema is defined `Replace[Name]`. This removes the need to pass things like id and created/updated at in these cases. Also updated the frontend connection accordingly.
* A lot of code quality and readability changes.
* Added a REPLACE for prices since it wasn't there yet.
* Removed the use of Upsert in Production since we don't use it elsewhere and thus don't need it.
* Fixed a bug where frontend tests would appear cached when they were indeed changed and should be running locally.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

Go to Swagger and look if all endpoints still work. The biggest changes here are just code so...

## ✅ CLOSED ISSUES
- Closes #256


backend/src/dto/dto.ts
```diff
@@ -48,38 +52,44 @@ import { createZodDto } from "nestjs-zod";
 export class ProductionDto extends createZodDto(ProductionSchema) {}
 export class ProductionViewDto extends createZodDto(ProductionViewSchema) {}
 export class CreateProductionDto extends createZodDto(CreateProductionSchema) {}
-export class UpdateProductionDto extends createZodDto(UpdateProductionSchema) {}
+export class ModifyProductionDto extends createZodDto(ModifyProductionSchema) {}
+export class ReplaceProductionDto extends createZodDto(
+  ReplaceProductionSchema,
+) {}
```
Nowak-Dominik, line 58: Maybe you can just write it on one line so it matches the style of other lines.
```suggestion
export class ReplaceProductionDto extends createZodDto(ReplaceProductionSchema) {}
```

backend/src/event/event.service.spec.ts
```diff
@@ -205,40 +204,25 @@ describe("EventService", () => {
       expect(dbService.updateEvent).toHaveBeenCalledWith(1, mockEvent);
       expect(result).toEqual(mockEvent);
     });
-
-    it("should throw BadRequestException if url id and body id do not match", async () => {
```
Nowak-Dominik, line 209: Why is there no need to test this anymore?
Is it becouse we only have the id in the url and not in the body anymore or something like that?

backend/src/production/controllers/production.controller.ts
```diff
@@ -119,30 +120,31 @@ export class ProductionController {
   @Put(":productionId")
   async replaceProduction(
     @Param("productionId", ParseIntPipe) productionId: number,
-    @Body(new ZodValidationPipe(ProductionSchema)) production: ProductionDto,
+    @Body(new ZodValidationPipe(ReplaceProductionSchema))
+    replaceProduction: ReplaceProductionDto,
```
Nowak-Dominik, line 124: The same as before, I know that this is purely details but maybe you can also just let it be on the same line.
However I could imagine that you palced the `replaceProduction: ReplaceProductionDto,` underneath becouse the line was getting quite long. So if that was your reason then fine ofcourse. You can leave it as it is.
```suggestion
    @Body(new ZodValidationPipe(ReplaceProductionSchema)) replaceProduction: ReplaceProductionDto,
```

Nowak-Dominik (approved):
Great job, @Alikznollet!
I reviewed everything, and it all looks well done to me, so I’m approving this PR.

I did leave a few minor suggestions related to code style. They’re completely optional, so feel free to apply them only if you agree, they’re not required for approval.

I also left a small question, just to make sure I understand something correctly.

backend/src/dto/dto.ts
Alikznollet: Problem with this is that Prettier doesn't like it and thus becomes a red warning line... 😢

backend/src/event/event.service.spec.ts
Alikznollet: Yeah all IDs are now passed through the URL, eliminating the need to check for these things.

backend/src/dto/dto.ts
Nowak-Dominik: Ah, ok haha then it is fine.

backend/src/production/controllers/production.controller.ts
Alikznollet: Same thing like the first comment, Prettier is quite litteraly forcing me to do this because of formatting 😢

Nowak-Dominik: Ok yeah, fine, no problem.

backend/src/event/event.service.spec.ts
Nowak-Dominik: Ok yeah I was thinking that it was becouse of that. Thanks for the explanation!

MrLegitFTW (approved):
I like this change for consistency & readability as discussed on monday.

approved

PR #269, opened 2026-03-26
Media on server

### LINKED ISSUE #254: Uploading/Serving/Storing of Media
* Work out a way to store and upload media to the backend
* Make Nginx or Nest serve these images on the server so we can use them.

### LINKED ISSUE #261: Document how we store/serve media.
Put this in the github pages.

MrLegitFTW:
## 🎯 MAIN GOAL
* adds media storage functionality + endpoints

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. try uploading through the web :-)

## ⚠️ REMAINING ISSUES

## ✅ CLOSED ISSUES
- #254


Alikznollet (commented):
At first I was going to write that I didn't understand it really, but after I read the docs this is very clear now.

It's very nice to see both a develop solution and prod solution where we can freely just use the same URL everywhere!

The URL being in the body of a POST seems a little smelly to me so we'll have to see about how to maybe solve that.

backend/src/media/media_storage/media_storage.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+  })
+  @ApiCreatedResponse({ description: "Saved media, returns the URL." })
+  @UseInterceptors(FileInterceptor("file"))
+  @Post()
+  async saveMedia(
+    @Body("url") url: string,
+    @UploadedFile() file: Express.Multer.File,
+  ): Promise<string> {
+    return await this.mediaStorageService.saveMedia(url, file.buffer);
+  }
```
landgeet, line 115: How should the url look like that you give this endpoint? I tried it in the swagger docs and kept getting 403 errors.

backend/src/media/media_storage/media_storage.service.ts
landgeet, line 1: I like how you did the storage here, but would it be possible to put these classes in different files? I just think it's bad practice to define multiple (non-record) classes in one file. But it's up to you if you do or don't think it's necessary.

landgeet (commented):
This looks great!
The code is overal well structured and is also seems to work correctly.
The docs are really clear and clear up most unclarities.
Just two remarks about structure and a usecase below:

backend/src/media/media_storage/media_storage.controller.ts
MrLegitFTW: should be an url for our server thus starting with https://sel2-4.ugent.be or just the url from the MEDIA_BASE_URL env var if you have a diff one set up.

backend/src/media/media_storage/media_storage.service.ts
MrLegitFTW: yes sure we can do that.

MrLegitFTW:
note: this should in theory also work for prints, not only photos. Although there are in dev no constraints on where you save this, on prod you will be forced to save prints under BASE_URL/prints and photos under BASE_URL/photos. Feel free to try this out :-)

backend/src/media/media_storage/service/local_media_storage.ts
Alikznollet, line 1: I would maybe not call this `service` but `local_media.storage.ts` or something like that?

backend/src/media/media_storage/media_storage.controller.ts
```diff
@@ -0,0 +1,82 @@
...
+   * @returns The media as a buffer.
+   * note: done like this, although weird, to allow having a body in the request which GET does not allow.
+   */
+  @ApiOperation({ summary: "Fetch media from a given URL." })
+  @ApiBody({ schema: { properties: { url: { type: "string" } } } })
+  @ApiOkResponse({ description: "Found media." })
+  @Post("fetch")
+  async getMedia(@Body() body: { url: string }): Promise<Buffer> {
+    return await this.mediaStorageService.getMedia(body.url);
+  }
```
Alikznollet, line 86: I've found a way to rewrite this to a GET (which would be a lot more REST.)

```ts
@Get("fetch")
async getMedia(
  @Query("url") url: string,
): Promise<Buffer> {
  // Optioneel validate de url.
  return await this.mediaStorageService.getMedia(url);
}
```

We kunnen een URL in de query sturen in de frontend met behulp van volgende lijnen:

```ts
const targetUrl = "https://example.com/my-image.jpg?size=large";
const fetchUrl = `/media/storage/fetch?url=${encodeURIComponent(targetUrl)}`;
```

Dit zal ervoor zorgen dat de data die url correct meegegeven wordt en Nest kan hier volledig mee omgaan.

backend/src/media/media_storage/service/remote_media_storage.ts
Alikznollet, line 1: Same here, maybe change this to `remote_media.storage.ts`?

backend/src/media/media_storage/media_storage.controller.ts
Alikznollet: Maybe document this somewhere if not already done?

backend/src/media/services/media.crop.service.ts
```diff
@@ -40,10 +49,48 @@ export class MediaCropService {
...
+    ) {
+      try {
+        // fetch the photo
+        const response = await fetch(createCrop.url);
+        const buffer = Buffer.from(await response.arrayBuffer());
+
+        // concat new url.
+        const parsedUrl = new URL(createCrop.url);
+        const ext = path.extname(parsedUrl.pathname) || ".jpg";
+        const newFileName = `${savedCrop.id}${ext}`;
```
Alikznollet: Maybe change to ´${savedCrop.id}-${savedCrop.name}${ext}´ to be more in line with what the scraper brings in.

Alikznollet (changes requested):
This looks strong so far. I've suggested a way to fix the POST problem and some other smaller things.

Are there seperate issues for the media storage connection with the frontend? Because those composables still need to be written.

backend/src/media/media_storage/service/local_media_storage.ts
MrLegitFTW: yes that makes sense

backend/src/media/services/media.crop.service.ts
MrLegitFTW: sure

backend/src/media/media_storage/media_storage.controller.ts
MrLegitFTW: I will give this a try and see if it has the desired effect.

backend/src/media/media_storage/media_storage.controller.ts
MrLegitFTW: this is documented in github pages under api/media_storage

backend/src/media/media_storage/media_storage.controller.ts
```diff
@@ -24,17 +36,23 @@ export class MediaStorageController {
...
   @ApiOkResponse({ description: "Found media." })
-  @Post("fetch")
-  async getMedia(@Body() body: { url: string }): Promise<Buffer> {
-    return await this.mediaStorageService.getMedia(body.url);
+  @Get("fetch")
+  async getMedia(@Query("url") url: string): Promise<StreamableFile> {
+    const buffer = await this.mediaStorageService.getMedia(url);
+
+    // return a file so this way we can actually see the file in swagger.
+    return new StreamableFile(buffer);
```
Alikznollet: Okay the StreamableFile thing is fire, I'll look at this later when on my PC with the downloaded media.

Alikznollet (commented):
Will approve later when I've been able to test the GET, other than that this is great. Expect an approval in about an hour.

Alikznollet (approved):
It works but it does return something rather weird... When you rename the file to `x.jpg` it does correctly open the image though.

backend/src/media/media_storage/media_storage.controller.ts
```diff
@@ -0,0 +1,100 @@
...
+    description: "The URL of the media (must be URL-encoded on the frontend)",
+  })
+  @ApiOkResponse({ description: "Found media." })
+  @Get("fetch")
+  async getMedia(@Query("url") url: string): Promise<StreamableFile> {
+    const buffer = await this.mediaStorageService.getMedia(url);
+
+    // return a file so this way we can actually see the file in swagger.
+    return new StreamableFile(buffer);
+  }
```
Alikznollet, line 86:
```suggestion
  async getMedia(@Query("url") url: string): Promise<StreamableFile> {
    const buffer = await this.mediaStorageService.getMedia(url);

    let contentType: string;
    const parsedUrl = new URL(url);
    const filename = path.basename(parsedUrl.pathname);
    const ext = path.extname(parsedUrl.pathname);
    switch (ext) {
      case ".jpg":
      case ".jpeg":
        contentType = "image/jpeg";
        break;
      case ".png":
        contentType = "image/png";
        break;
      case ".gif":
        contentType = "image/gif";
        break;
      case ".webp":
        contentType = "image/webp";
        break;
      case ".pdf":
        contentType = "application/pdf";
        break;
      default:
        contentType = "application/octet-stream";
        break;
    }

    // return a file so this way we can actually see the file in swagger.
    return new StreamableFile(buffer, {
      type: contentType,
      disposition: `attachment: filename="${filename}"`,
    });
  }
```

This fixes the file extension issue! You'll have to import `node:path` at the top! When you ask for an image now the browser knows what it's downloading.

Alikznollet (changes requested):
Fixed the issue I posted in my previous review.

backend/src/media/media_storage/media_storage.controller.ts
MrLegitFTW: that's great, thanks for this.

Alikznollet (approved):

landgeet (approved):
looks good now!

PR #286, opened 2026-03-28
Prints db update

MrLegitFTW:
## 🎯 MAIN GOAL
- [x] this PR implements a new feature
* updates the db and media so that prints can be implemented
- [ ] this PR fixes some existing bugs

## 🛠️ TESTING
- [ ] there is enough test coverage for this code
- [x] all tests succeed

## 📝 DOCUMENTATION
- [x] there is enough documentation written for this code
- [x] external documentation (README, etc.) has been changed

## 🔍 HOW TO TEST
1.  same as always

## ⚠️ REMAINING ISSUES
- [ ] there are still remaining bugs in the code
- [ ] this code requires changes elsewhere (changed database structure, API-endpoint, etc.)

## ✅ CLOSED ISSUES
- #218 

## NOTE:
This will not impl the db service or enpoints for prints as these are seperate issues. nor the tests for media or storage of media/prints proper.

PR #289, opened 2026-03-28
Media Scraper FR

### LINKED ISSUE #275: Upgrade Scraper for Media
- [ ] Update the scraper to also scrape media.
- [ ] Make it upload the media.

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* Media scraping support for the scraper service.
  * Now scrapes galleries, items and crops we need from the vnv API. 
  * A LOT of readability improvements regarding scraping progress and insertion progress.

* Changed the formats we store for crops to newer formats.
* General Scraper improvements regarding speed and batching that enable us to more safely scrape data.
* Automatic fetching of images in batches.
* Reset Script (and example CSV) follow the `command` pattern for Nest that allows it to use modules without being part of the actual app!
* CVS Injector fully made Nest compatible!!!

## 🛠️ TESTING

We ain't testing this :-)

## 📝 DOCUMENTATION

**Missing documentation:**
* Moved to #298 

## 🔍 HOW TO TEST

Set the `ENABLE_SCRAPER` env variable to `true` and see the magic work when you turn on the backend.

## ⚠️ REMAINING ISSUES

This PR relies on the media functionality to be **stable**. That means I will wait for #269 and #286  to be merged before even considering merging this.

There's also still a lot of *dummy* values in the code right now that I still have to clean up.

## Todo list

- [x] Make DB Connection work with TRANSACTIONS (and maybe even batched).
- [x] Add Insertion functionality for Gallery, Item and Crop. 
- [x] Automatic batch downloading of media. + Progress via COUNT query

## ✅ CLOSED ISSUES
- Closes #275 

Already implemented part of #297 but will be doing the rest in seperate PR later.


Alikznollet:
This PR is ready for review!

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
...
+    const parsedEvents = await CSVFileParser.parseEventsCSV(input);
+
+    // deduplicate locations from events and convert to csvLocations
+    const locationByName = new Map<string, vnvLocation>();
+    for (const row of parsedEvents) {
+      const key = `${row.location.en}||${row.location.nl}`;
+      if (!locationByName.has(key)) {
+        locationByName.set(key, this.toCsvLocation(row.location));
+      }
+    }
```
MatthieuDC, line 213: Is there a way to maybe create Batch operations for this instead of using for loops? This will be a lot better for large CSV's

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
+/* eslint-disable @typescript-eslint/no-unsafe-assignment */
+/* eslint-disable @typescript-eslint/no-unsafe-member-access */
```
MatthieuDC, line 2: Disabling ESLint seems a bit odd, defeating the purpose of TypeScript. You could create some simple interfaces to strongly type the parsed CSV rows instead of using any

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
...
+    description: production.description1,
+    description_2: production.description2 ?? { en: "N/A", nl: "N/A" },
+    info: production.credits ?? { en: "N/A", nl: "N/A" },
+    events: [],
+    genres: tagsForProduction.map((t) => `csv-${t}`),
+    galleryId: "",
+  };
+}
+
+function toOldCsvEvent(
```
MatthieuDC, line 60: You have a lot of repeated logic between your OldCsv and normal Csv functions. Maybe it could be better to have one function that behaves differently based on a "tags?: string[]" (for example) that defines the difference? Just brainstorming here

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
...
+          this.toCsvLegacyIdFromNumericId(productionId);
+        try {
+          const production =
+            await this.dbConnection.getProductionByLegacyId(productionLegacyId);
+          await this.dbConnection.linkTag(production.id, tagId);
+        } catch {
+          this.logger.warn(
+            `Skipping tag link; production not found for ${productionLegacyId}`,
+          );
+        }
```
MatthieuDC, line 274: Wouldn't it be better to specifically catch something here (like an error type or anything?) so it's a bit clearer where the exception happens? Again just an idea, see what is plausible

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
...
+        );
+      }
+    }
+
+    this.logger.log("Structured prices CSV injection completed.");
+  }
+
+  async injectOldCsvData() {
+    const productionsFile = "../common/res/productions_output.csv";
+    const eventsFile = "../common/res/events_voorstellingen.csv";
```
MatthieuDC, line 355: Is it possible to use environment variables here instead of hardcoded variables?

backend/src/util/scraper/csv/inject-csv.engine.ts
```diff
@@ -0,0 +1,406 @@
...
+
+    const csvTags: vnvGenre[] = tags.map(toOldCsvTag);
+    const csvProductions: vnvProduction[] = productions.map((production) =>
+      toOldCsvProduction(
+        production,
+        productionTagLinks
+          .filter((link) => link.legacyId === (production.legacy_id ?? ""))
+          .map((link) => link.tagName),
+      ),
+    );
```
MatthieuDC, line 370: running .filter() on the entire array may lead to an O(N x M) complexity here, which could become problematic. Maybe you can make a dict/map of tags before the loop?

MatthieuDC (changes requested):
Overall a very nice service, but I've added some possible changes here. Check them out and have a look if these changes are possible (or if I'm rambling out of my mind)

Alikznollet:
> Overall a very nice service, but I've added some possible changes here. Check them out and have a look if these changes are possible (or if I'm rambling out of my mind)

All of these are things regarding @landgeet sooo, I just refactored these to work with the new services.

backend/src/util/scraper/database/db.connection.ts
```diff
@@ -515,8 +744,285 @@ export class UtilsDbConnection {
...
+    const query = `
+      INSERT INTO media_crop (
+        legacy_id, name, url
+      )
+      VALUES ($1, $2, $3)
+      ON CONFLICT (legacy_id)
+      DO UPDATE SET
+        name = EXCLUDED.name,
+        url = EXCLUDED.url
+      RETURNING *;
```
MrLegitFTW, line 776: maybe don't use `RETURNING *` seeing as that implies you need all fields (including legacy_id which we don't use in our objects). If you don't need anything then don't return anything and if you just need a check if it worked then you could use `RETURNING id`.

backend/src/util/scraper/database/db.connection.ts
```diff
@@ -515,8 +744,285 @@ export class UtilsDbConnection {
...
+      DO UPDATE SET
+        type = EXCLUDED.type,
+        original_filename = EXCLUDED.original_filename,
+        position = EXCLUDED.position,
+        width = EXCLUDED.width,
+        height = EXCLUDED.height,
+        credits = EXCLUDED.credits,
+        description = EXCLUDED.description,
+        title = EXCLUDED.title
+      RETURNING *;
```
MrLegitFTW, line 814: same here

backend/src/util/scraper/database/db.connection.ts
```diff
@@ -515,8 +744,285 @@ export class UtilsDbConnection {
...
+    const query = `
+      INSERT INTO media_gallery (
+        legacy_id, name, type
+      )
+      VALUES ($1, $2, $3)
+      ON CONFLICT (legacy_id)
+      DO UPDATE SET
+        name = EXCLUDED.name,
+        type = EXCLUDED.type
+      RETURNING *;
```
MrLegitFTW, line 907: here

backend/src/util/scraper/database/db.connection.ts
```diff
@@ -515,8 +744,285 @@ export class UtilsDbConnection {
...
+    galleryId: number,
+    itemId: number,
+    client?: PoolClient,
+  ): Promise<boolean> {
+    const rows = await this.query(
+      `
+        INSERT INTO gallery_item (gallery_id, item_id)
+        VALUES ($1, $2)
+        ON CONFLICT DO NOTHING
+        RETURNING *;
```
MrLegitFTW, line 967: And all other places basically. It's fine to leave this if you want however I think it makes it clearer to what the code actually needs to do?

backend/src/util/scraper/scraper.engine.ts
```diff
@@ -0,0 +1,339 @@
...
+import { Injectable } from "@nestjs/common";
+import { AppLogger } from "../logger/logger.service";
+import Bottleneck from "bottleneck";
+import { CROP_NAMES, CropName } from "@repo/common/src/objects/media";
+import { ConfigService } from "@nestjs/config";
+
+/**
+ * The Base of the VNV API.
+ */
+const apiBase: string = "https://www.viernulvier.gent";
```
MrLegitFTW: Perhaps an env var instead?

backend/src/util/scraper/scraper.service.ts
```diff
@@ -97,4 +102,100 @@ export class ScraperService implements OnApplicationBootstrap {
...
   }
+
+  /**
+   * Image Scraping Job
+   */
+
+  private isProcessingMedia = false;
+  private readonly downloadAmount = 50;
+
+  @Cron(CronExpression.EVERY_MINUTE)
```
MrLegitFTW, line 120: after thinking I think it might be better to just run this while at down-time like 3-am?
that way the cpu's limited resources can be used fully by the application instead of being split? Perhaps this is something we will see when we test though.

backend/src/util/scraper/database/db.connection.ts
```diff
@@ -25,22 +39,35 @@ import { Injectable } from "@nestjs/common";
...
   private pool: Pool;
-  constructor() {
+  constructor(private readonly configService: ConfigService) {
     this.pool = new Pool({
-      user: process.env.DB_USER_DEV,
-      host: process.env.DB_HOST_DEV,
-      database: process.env.DB_NAME_DEV,
-      password: process.env.DB_PASSWORD_DEV,
-      port: Number(process.env.DB_PORT_DEV),
+      user: this.configService.get<string>("DB_USER_DEV"),
```
MrLegitFTW, line 56: is there a reason for using configService instead of process.env? If so it might be better to use it everywhere for code consistency?

backend/src/util/scraper/scraper.service.ts
```diff
@@ -97,4 +102,100 @@ export class ScraperService implements OnApplicationBootstrap {
...
+  async processImages() {
```
MrLegitFTW, line 121: Is there a way at the moment we can keep the current urls if so desired and thus not auto-download?

MrLegitFTW (commented):
Overal very solid. I have some questions/ suggestions based mostly on personal preference feel free to check them out and let me know what you plan on doing. Besides that a few extra questions:
- Based on what @MatthieuDC said about the any typing. It is my understanding we do this seeing is the returned objects after parsing are still any for generic reasons, therefore we still need to map them first. This seems okay to keep like this or are there other plans/ architecture visions for this?
- Have the docs been updated with the new changes? Or is there a seperate issue for this?
- Testing wise I am guessing and also suggesting we wait for a test deployment so we can actively see the impact.
- Is there a way (looking in the future) we could re-structure this a bit to make it a bit easier to navigate ?

backend/src/util/scraper/scraper.engine.ts
Alikznollet: This could most definitely be an env var, will change this.

backend/src/util/scraper/scraper.service.ts
Alikznollet: Thing is, when it is running with higher limits (e.g. 200 per 10 secs) this could bombard both servers and we don't really know clearly if there are API limits. We could start by testing out higher limits though.

backend/src/util/scraper/database/db.connection.ts
Alikznollet: Already made an issue #297 for this, was thinking of doing that seperately after this was done.

backend/src/util/scraper/scraper.service.ts
Alikznollet: yeah with an env var, will fix.

backend/src/util/scraper/database/db.connection.ts
Alikznollet: I'll just respond to this:

I'll change this when rewriting the scraper if that's okay.

backend/src/util/scraper/database/db.connection.ts
MrLegitFTW: that works this pr is already quite big

backend/src/util/scraper/scraper.service.ts
MrLegitFTW: I see, perhaps we should tests the limits thursday?

backend/src/util/scraper/database/db.connection.ts
MrLegitFTW: Sure it works now and in theory there is no reason to change it.

backend/src/util/scraper/scraper.service.ts
Alikznollet: Sounds like a plan.

Alikznollet:
@MrLegitFTW Docs are also marked as a seperate issue by the way.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: We only use these once, but maybe for security measures we can do this? Even with people possibly seeing the relative path i don't really think it's needed, but can do.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: I didn't add this so @Alikznollet has to answer here :)

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: I could do this, but i wanted to have most of the old and new stuff seperatly because the new data might change in the future and it would be easier to change more freely.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: Do you mean something like this?:
```suggestion
        } catch (error) {
          if (error instanceof ResourceGoneException) {
            this.logger.warn(
              `Skipping tag link; production not found for ${productionLegacyId}`,
            );
            continue;
          }
  
          this.logger.error(
            `Unexpected error while linking tag to production ${productionLegacyId}`,
            error,
          );
          continue;
        }
```
I won't be throwing the errors because i don't want the parser to stop if one link doesn't work, so i just continue.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: Yes this is a good call, i will fix this.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: Do you only mean this code or in general? Because if you mean in general, than i would have to write some new functions for this. And i think the csv files that they will import in the future won't be longer then the old csv file (100.000+ lines) so i would maybe do this if we have time leftover for milestone 3.
If you mean only this code, i don't really understand what you mean here. Do you mean something like this?:
```suggestion
    const locationByName = new Map<string, vnvLocation>(
      parsedEvents.map((row) => [
        `${row.location.en}||${row.location.nl}`,
        this.toCsvLocation(row.location),
      ]),
    );
```

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: I just tried and it would result in more code than before so i don't think it's useful to do this in my opinion.

backend/src/util/scraper/csv/inject-csv.engine.ts
landgeet: I don't think it is really worth doing this after trying because it just make the inject engine involve the configservice which seems unnecessary in my opinion.

MrLegitFTW (approved):
Going to approve this, we will be refactoring this for sure, but for now lets get testing please. :-)

PR #290, opened 2026-03-29
Language selector header

### LINKED ISSUE #279: Language selector header

jillvanham:
## 🎯 MAIN GOAL
- [ ] this PR implements the new language selector
* bug in header fixed: header now always visible after resizing screen

## 🛠️ TESTING
**Missing tests:**
* all

## 🔍 HOW TO TEST
1. npm run start:dev or with the preview link
3. check the new language local selector 

## ✅ CLOSED ISSUES
- #279

### DIFF
--- a/frontend/app/components/LocaleSelector.vue
+++ b/frontend/app/components/LocaleSelector.vue
@@ -13,8 +13,22 @@ const closeDropdown = (e) => {
   }
 }
 
-onMounted(() => window.addEventListener('click', closeDropdown))
-onUnmounted(() => window.removeEventListener('click', closeDropdown))
+// Close when 'Escape'
+const handleEscape = (e) => {
+  if (e.key === 'Escape') {
+    isOpen.value = false
+  }
+}
+
+onMounted(() => {
+  window.addEventListener('click', closeDropdown)
+  window.addEventListener('keydown', handleEscape)
+})
+
+onUnmounted(() => {
+  window.removeEventListener('click', closeDropdown)
+  window.removeEventListener('keydown', handleEscape)
+})
 
 const handleLocaleChange = (code) => {
   setLocale(code)
--- a/frontend/app/components/LocaleSelector.vue
+++ b/frontend/app/components/LocaleSelector.vue
@@ -1,10 +1,14 @@
 <script setup>
 import { ref, onMounted, onUnmounted } from 'vue'
-import { ChevronDown } from 'lucide-vue-next'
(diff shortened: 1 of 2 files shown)


frontend/app/components/LocaleSelector.vue
```diff
@@ -0,0 +1,72 @@
...
+      leave-to-class="transform opacity-0 scale-95"
+    >
+      <div
+        v-if="isOpen"
+        class="absolute right-0 z-[110] mt-2 w-full origin-top-right rounded-md border-2 border-[var(--foreground)] bg-[var(--background)] shadow-lg outline-none"
+      >
+        <div class="py-1">
+          <button
+            v-for="loc in locales"
+            :key="typeof loc === 'string' ? loc : loc.code"
```
landgeet: I would add a variable that stores `typeof loc === 'string' ? loc : loc.code` like this:
`const getCode = (loc) => typeof loc === 'string' ? loc : loc.code`.
And this line would look like this:
`:key="getCode(loc)"`
This would make the code more readable and avoid code duplication.

landgeet (changes requested):
This looks really clean.
There is a small nitpicky remark below and just a little question:
Would it be possible to make the dropdown closable with the 'Escape' key?

Robindr145:
The purple text seems a bit off for me. In my opinion these are to much colours in 1 space. Maybe a default black text with a clear selector (blue check mark) would do the job (on light mode). It's not clear which language is selected atm.

Maybe this is just preference. I don't know what the others think.
Schermafbeelding 2026-03-29 211326

ddndauw:
As far as I know, the functionality of the button works perfect. The translations work, the different screen sizes work, etc.
As for the layout, I do have double feelings about the purple color, the one Robin mentioned. At first it seems kind of out of place, but the longer I look at it, the more I get used to it? So for me, the purple color can stay if you want to.

aaa

I am not really a fan of the hover though, but I cannot place exactly what I find off about it. Maybe it's too dark? Or maybe the little white space on top and below when you hover over something kinda ick me. But these are all minor details.

Great work otherwise! The header still looks very amazing.

ddndauw:
Code wise I do not really have any other remark. 👍

Alikznollet (approved):
This looks great! No remarks from me.

ddndauw (approved):
Thank you for changing the language dropdown, it looks better now! I have no further remarks, good job !

PR #293, opened 2026-03-29
Production story component

ddndauw:
## 🎯 MAIN GOAL
- [x] this PR implements a component that displays the stories that are linked to the production

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Run `npm run start:dev`
2. Go to `http://localhost:3001/tempShowcase`

First version is a version with 3 stories. Second version has no stories. The last version has more than 3 stories, and is scrollable.

Please check light/dark mode, the language button and different screen sizes.
Check if the new test (+ all previous tests) succeed.

## ⚠️ REMAINING ISSUES
\-

\-

## ✅ CLOSED ISSUES
- #282


jillvanham (commented):
Great base again! The palette is a bit too neutral for me (maybe I'll look into that when I'm integrating this component into the detail page), but otherwise, it's looking good. I was just wondering if it would be better to reuse the 'card' design from the stories page here? It feels like a waste of the effort you've put into these new ones, so it's just a suggestion, but it might be worth thinking about. What do you think?

Alikznollet (commented):
This looks great, I suppose `StoryItem` will be swapped out by actual fetched blogs when the time comes?

ddndauw:
@jillvanham I do think Robin's code is reusable, I'll look into it! 

@Alikznollet I originally created `StoryItem` as some sort of "frontend DTO" since it only consists of the parts that are needed for the component, so that if something in the backend would change, only the mappings to `StoryItem` would have to be changed. But I just realised that would probably never happpen/wouldn't be an actual issue................................. And the fields are also practically the same already....................................................... It's probably better if I directly use the object defined in common... I'll do that, thank you for opening my eyes.

Alikznollet:
> @jillvanham I do think Robin's code is reusable, I'll look into it! 
> 
> @Alikznollet I originally created `StoryItem` as some sort of "frontend DTO" since it only consists of the parts that are needed for the component, so that if something in the backend would change, only the mappings to `StoryItem` would have to be changed. But I just realised that would probably never happpen/wouldn't be an actual issue................................. And the fields are also practically the same already....................................................... It's probably better if I directly use the object defined in common... I'll do that, thank you for opening my eyes.   

No worries! The objects in common should automatically contain the same things from the backend at all times even if they're changed. The image for example is something you'll have to fetch seperately when that is eventually implemented.

ddndauw:
@jillvanham I have updated it to reuse Robin's component to avoid duplication.
@Alikznollet I have removed `StoryItem`.

frontend/app/components/production/storyListView.vue
```diff
@@ -0,0 +1,70 @@
...
+interface Props {
+  stories: BlogView[];
+}
+const props = defineProps<Props>();
+
+const sortedStories = computed(() => // function to sort the stories, oldest first
+    [...props.stories].sort(
+        (a, b) => new Date(a.created_at).getTime() - new Date(b.created_at).getTime()
+    )
+);
```
Alikznollet: I think Blogs are already sorted by date when they are pulled from the backend. If that's not the case maybe we should look into making an issue that adds that functionality (so the frontend doesn't have to do any sorting).

Alikznollet (approved):
This looks great visually and code wise. One small remark that could maybe lead to one small change too. I don't think it matters *that* much but maybe good for consistency if we keep all sorting out of the frontend.

ddndauw:
@Alikznollet I removed the sorting 👍

jillvanham (approved):

PR #304, opened 2026-03-31
Production detail page

### LINKED ISSUE #280: Production detail page
displays:
- [ ]  image
- [ ] title
- [ ] tags
- [ ] description
- [x] #281 
- [x] #282 
- [x] #283

jillvanham:
## 🎯 MAIN GOAL
- [ ] this PR implements the detail page of a production

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* Unit tests for the `cleanText` helper and `isValid` utility

## 📝 DOCUMENTATION

**Missing documentation:**

## 🔍 HOW TO TEST
1. Run `npm run start:dev` 
2. Check `http://localhost:3001/tempShowcase`
3. Navigate to `http://localhost:3001/productions/[id]` or select a production from the homepage.

## ⚠️ REMAINING ISSUES
* media not yet implemented 
* missing: table for prints

## ✅ CLOSED ISSUES
- #280


ddndauw:
I have looked at your code visually-wise, and I must say, good job for the moment! I especially like the dark mode. Here are some few remarks you can still change, some a bit more nitpicky than others:

- I don't know if it is something you can fix, but right now, when I hover my mouse over the event table, it changes from a normal arrow to a hand, even though there's nothing clickable. I cannot really take a screen of it since it doesn't screen my mouse haha, but I can send a picture on Discord if you don't understand what I mean. This doesn't happen with the story list view for example, so I am wondering if there is some fix to it.
- I think I like the credits a little bit more centered? In the sense that, there are now 2 columns, but they do not take up the space that there is. There is like a lot of space on the right but the space isn't big enough for a third column.
- I have not checked the code yet, but did you cut off based on an amount of characters, or on how big the section gets. If you did the first option, maybe you can try cutting off once a certain height is reached? Right now for example, on small screens, it gets cut off at the same word, but like it's still the same line, it feels a bit counter-intuitive. But if it's not possible it's also not that much of a big deal.

desc

- Maybe try to implement the "..." too for when the title gets too long, same with the artist, description2 and tagline. The real data probably won't be that extreme, but you never know. I didn't immediately see any other places with this problem but feel free to experiment with length yourself a bit ! 

ooo
ooo2

- This part is not that visible in light mode, but is very visible in dark. So maybe make it a bit more noticable in light mode ? (The background color)

ddd

- xevenementen
xevenementen

frontend/app/pages/productions/[id].vue
```diff
@@ -0,0 +1,262 @@
+<script setup lang="ts">
+import { ref, computed } from 'vue';
+import { ChevronLeft } from 'lucide-vue-next';
+import type { ProductionView, TagView } from "@repo/common";
+import ProductionStories from "../../components/production/storyListView.vue";
+import EventTable from "../../components/production/EventTable.vue";
```
Comment (ddndauw) [inline frontend/app/pages/productions/[id].vue]: You can use Nuxt's auto-imports for these by using ProductionEventTable and ProductionStoryListView as component names.

frontend/app/components/production/EventTable.vue
```diff
@@ -11,155 +11,219 @@
...
+    minute: "2-digit",
+  });
+};
+
+const formatPrice = (price: number) => {
+  return new Intl.NumberFormat(locale.value, {
+    style: "currency",
+    currency: "EUR",
+  }).format(price);
+};
```
landgeet, line 80: These functions already exist in the `formatters.ts` file.
All formatter functions are defined in there and you can always add some if necessary.

frontend/app/components/production/storyListView.vue
landgeet, line 1: why did you remove these things?

frontend/app/components/production/EventTable.vue
landgeet, line 1: I don't really like the chevron with show more/less for this table. I don't think there will ever be that many events for one production that this will be necessary.

I think it would be nice to have start-end time instead of just the starttime.

I do really like the extra prices button that shows all addtional prices!

landgeet (changes requested):
Visually this looks great!
I haven't really looked at the [id].vue code yet (i will do this later), but i have at the rest of the code.
The remarks on this code you can find below.
I have found a bug in the 'back' button on the id page:
When going to a different page with the arrows, click a production and then go back, the page counter still shows the previous page number but it shows the first page. If you then use the arrows it works again. This probably has to do with a resetting/reloading mechanism. See picture below for an example:

Go to page 8:
Image
Click the production:
Image
Go back, now the production are from page 1 but number 8 still shows:
Image
Use the back arrow to go to page 7 and now the correct data shows again:
Image

frontend/app/components/production/storyListView.vue
jillvanham: I moved that logic into the parent detail page for a few reasons: 
- Handling the 'Stories' title in the parent ensures that all headers follow the same design convention. 
- I check if there are stories in the parent (by checking length). Now this component only cares about rendering the list you give it, while the parent decides if and where that section should appear based on the data.

frontend/app/components/production/EventTable.vue
jillvanham: Thanks for pointing that out! I'm planning to clean this up in a new issue where I'll also be improving the events table.

frontend/app/components/production/EventTable.vue
jillvanham: I've increased the default limit to 5 events to make the 'show more' button less likely to appear. I'll address the start/end time and further table improvements in a new issue so we can merge this one.

Alikznollet (approved):
This looks great! Don't forget to remove the tempShowcase page before you merge.

frontend/app/pages/archive/[id].vue
```diff
@@ -0,0 +1,330 @@
...
+);
+
+/**
+ * Check if a string is useful (not "N/A" or empty)
+ */
+const isValid = (val: any) => {
+  if (!val) return false;
+  const s = String(val).trim().toUpperCase();
+  return (
+    s !== "" && s !== "N/A" && s !== "UNDEFINED" && s !== "\\N" && s !== "\N"
```
Comment (landgeet) [inline frontend/app/pages/archive/[id].vue]: just a small warning here: `s !== "\N"` gets interpreted by typescript as `s !== "N"` (backslash gets ignored) so this could lead to bugs where a string `"N"` isn't valid.

landgeet (changes requested):
Thanks for changing my previous remarks.
This looks great now and is almost ready to be merged!
I just have a really small remark below about the `[id].vue` page when i was skimming over it.
If this is resolved, this is ready to be merged!

frontend/app/pages/archive/[id].vue
```diff
@@ -0,0 +1,330 @@
...
+const { t, locale } = useI18n();
+const router = useRouter();
+const { getById, getTags, getBlogs } = useProductionApi();
+const route = useRoute();
+
+const goBack = () => {
+  if (window.history.length > 1) {
+    router.back();
+  } else {
+    router.push("/archive"); // Fallback
```
Comment (Nowak-Dominik) [inline frontend/app/pages/archive/[id].vue]: use the archive.base from routes file

Nowak-Dominik (approved):
@jillvanham 

This looks good to me, only change the one thing that I pointed out please.
After that you can merge it!

PR #320, opened 2026-04-04
working calendar

Robindr145:
## 🎯 MAIN GOAL
 Implements the genaral calendar files + visible on the blogs-page

## 🛠️ TESTING
 none

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. run `npm run start:dev` and go to the /stories-section
2. go to the calendar icon and try it out.

## ✅ CLOSED ISSUES
- #320

### DIFF
--- a/frontend/app/components/DefaultCalendar.vue
+++ b/frontend/app/components/DefaultCalendar.vue
@@ -1,5 +1,5 @@
 <!--
-  components/calendar.vue
+  components/DefaultCalendar.vue
   =======================
   Orchestrates all sub-components. Owns all state; children are presentational.
   Sync flow:
--- a/frontend/app/components/DefaultCalendar.vue
+++ b/frontend/app/components/DefaultCalendar.vue
@@ -12,8 +12,9 @@ import CalendarTabs   from "~/components/calendar/CalendarTabs.vue";
 import CalendarInputs from "~/components/calendar/CalendarInputs.vue";
 import CalendarMonths from "~/components/calendar/CalendarMonths.vue";
 import CalendarFooter from "~/components/calendar/CalendarFooter.vue";
-import type { CalMode }    from "~/components/calendar/CalendarInputs.vue";
-import type { MonthData }  from "~/components/calendar/CalendarMonths.vue";
+import type { CalMode }   from "~/components/calendar/CalendarInputs.vue";
+import type { MonthData } from "~/components/calendar/CalendarMonths.vue";
+import { localIso, localTodayIso, buildWeekdayLabels } from "~/utils/formatters";
 
 interface DateFilter {
   after?:  string;
@@ -36,17 +37,6 @@ const { t, locale } = useI18n();
 
 const intlLocale = computed(() => locale.value === "en" ? "en-GB" : "nl-BE");
 
-// Timezone-safe helpers
-
-function localIso(d: Date): string {
-  const y  = d.getFullYear();
-  const m  = String(d.getMonth() + 1).padStart(2, "0");
-  const dd = String(d.getDate()).padStart(2, "0");
-  return `${y}-${m}-${dd}`;
-}
-
-const todayIso = computed(() => localIso(new Date()));
-
 // State
 
 const mode      = ref<CalMode>("range");
@@ -59,7 +49,7 @@ const today     = new Date();
 const viewYear  = ref(today.getFullYear());
 const viewMonth = ref(today.getMonth());
 
-// Sync props for CalendarInputs (calendar → text fields) 
+// Sync props for CalendarInputs (calendar → text fields)
 // These computed values mirror the current selection into the input components.
 
 const syncSingle = computed(() =>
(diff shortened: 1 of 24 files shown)


ddndauw:
The calender looks very cool and impressive already!! 
I haven't looked at it very thoroughly yet since you're still busy, but maybe a few small remarks for when you continue to work at it:
- I don't know if it's only my screen, but the icon and the text aren't centered (the icon is a bit lower).
button

- From what I understand "Until now" is a date you selected until today, and "Oldest" is everything before a selected date? I didn't immediately get what they meant, so I'm wondering if other labels would suit better? Maybe something like "From date" and "To date"?, I'm not sure though.
- I do not know if it's possible, but it would be cool if the calendar and the fields where we can type in were synced.

Robindr145:
In an upcoming issue, the filters (select on year) will be added to the blogs-page, so the date-button will be replaced with a filter button, where we also can find the calendar.

Nowak-Dominik:
@Robindr145
Not gonna lie, I’m speechless, this looks so, so good. Very clean! 🔥🔥🔥

I do have a couple of small suggestions:

Maybe you could make the border of the “Clear” button a bit more visible, so the button stands out more.

image

Also, I’m not sure exactly where it would fit best, but it might be nice to have a “Go to today” button that resets the calendar view to today’s date. Right now, that already happens when the user opens the calendar, but having a button for it would make it easier to quickly return to that view.

frontend/app/components/calendar/CalendarInputs.vue
```diff
@@ -0,0 +1,260 @@
...
+const errorStart  = ref(false);
+const errorEnd    = ref(false);
+
+// Format helpers
+
+/** Display ISO as DD/MM/YYYY to match the placeholder hint */
+function formatForInput(iso: string): string {
+  const [y, m, d] = iso.split("-");
+  return `${d}/${m}/${y}`;
+}
```
Nowak-Dominik: Maybe you can put this in `utils/formatters.ts`

frontend/app/components/calendar/CalendarMonths.vue
```diff
@@ -0,0 +1,90 @@
...
+      :is-range-start="isRangeStart"
+      :is-range-end="isRangeEnd"
+      :is-in-range="isInRange"
+      :is-today="isToday"
+      :is-future="isFuture"
+      :is-range-cap-left="isRangeCapLeft"
+      @click-day="(iso) => emit('click-day', iso)"
+      @hover-day="(iso) => emit('hover-day', iso)"
+      @leave-day="emit('leave-day')"
+    />
```
Nowak-Dominik, line 63: I don't know if that is important but I noticed that the right CalendarGrid underneath has a class property:
`class="cal-month--right"` so maybe the left one also needs it? This is just a suggestion, maybe it's unnecessary.

```suggestion
    <CalendarGrid
      class="cal-month--left"
      :label="leftMonth.label"
      :days="leftMonth.days"
      :weekdays="weekdays"
      :is-selected="isSelected"
      :is-range-start="isRangeStart"
      :is-range-end="isRangeEnd"
      :is-in-range="isInRange"
      :is-today="isToday"
      :is-future="isFuture"
      :is-range-cap-left="isRangeCapLeft"
      @click-day="(iso) => emit('click-day', iso)"
      @hover-day="(iso) => emit('hover-day', iso)"
      @leave-day="emit('leave-day')"
    />
```

frontend/app/components/DefaultCalendar.vue
```diff
@@ -0,0 +1,411 @@
...
+  return undefined;
+});
+
+// Month grid data 
+
+const rightYear     = computed(() => viewMonth.value === 11 ? viewYear.value + 1 : viewYear.value);
+const rightMonthIdx = computed(() => (viewMonth.value + 1) % 12);
+
+const weekdays = computed(() => {
+  const mon = new Date(2024, 0, 1); // known Monday
```
Nowak-Dominik: Can you explain why you need to work with a hardcoded known monday?

Nowak-Dominik (changes requested):
@Robindr145 
Very solid PR, great job!

I have a small thing that I asked to change, and a small question.
This PR is ready to be approved after you check the things I said and answer it.

frontend/app/components/calendar/CalendarInputs.vue
Robindr145: will move all ISO-related stuff in the `utils/formatters.ts`- file

frontend/app/components/calendar/CalendarMonths.vue
Robindr145: no extra class needed on the left grid. .cal-month is the base style; .cal-month--right is purely additive (it adds the left border and extra left padding). The left panel is already correct as-is.

frontend/app/components/DefaultCalendar.vue
Robindr145: The goal is to produce locale-aware single-character weekday labels (M, D, W… in Dutch; M, T, W… in English) starting from Monday, because the calendar uses a Monday-first week grid. toLocaleDateString with { weekday: "narrow" } gives you the right letter for any locale, but it needs an actual Date to format. It doesn't accept a day-of-week index directly. So rather than hardcoding ["M","T","W",…] per locale, the code picks a date that is known to be a Monday and iterates 7 consecutive days from it. January 1 2024 happens to be a Monday, so the 7-day loop lands exactly on Mon → Sun in the right order. The year doesn't matter at all; it's just an anchor.

Nowak-Dominik (approved):
@Robindr145 
Ok, thanks for the changes. Great job!

Alikznollet (approved):
This looks absolutely fire, code wise I didn't see any problems. If there are any that slipped past me we'll just fix them later. For now this looks very ready to merge!

Alikznollet:
@Robindr145 Also you'll probably have to fix the merge conflicts first, though they won't be very hard I'm sure.

PR #331, opened 2026-04-08
Refactor DB Services

### LINKED ISSUE #330: Cleanup DB Services
DB Services are very varied in their implementation

- [ ] Clean them up and make them the same
- [ ] Make them faster
- [ ] Fix any bugs on the way.

Alikznollet:
## 🎯 MAIN GOAL
- [x] this PR refactors some important code to make it more readable.

### Changelog

* All (outside of a couple specific things) DB Services have been refactored to use the new `db-utils.ts` helper functions.
  * This boosts the readability of the Database services a ton.
  * Less error prone because resulting clauses and updates and such are derived from the Zod objects in `common`.

* Added an AccountDto so that we don't have to return any anymore in the AccountDbService. This fixes some warnings and shouldn't change any functionality.

## 🛠️ TESTING

We're not testing DB services.

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

Run Swagger and test the endpoints whether everything still works.

## ✅ CLOSED ISSUES
- Closes #330


backend/src/database/media/db.media_crop.service.ts
```diff
@@ -80,30 +88,36 @@ export class MediaCropDatabaseService {
    * @returns The created crop.
    */
   async createCrop(crop: CreateMediaCropDto): Promise<MediaCropDto> {
-    if (!crop.name || !crop.url) {
-      throw new BadRequestException("Missing required fields");
-    }
```
MatthieuDC, line 85: I assume this is going in db-utils later? Or does something prevent this issue from occuring? Or is everything that just fails pushed further in the final error handler (if length == 0)?

MatthieuDC (commented):
Looks very good, only one comment about some error handling.

backend/src/database/media/db.media_crop.service.ts
Alikznollet: The CreateCropDto object will ALWAYS have name and url. The Schema is created so that these have to be in the object or the ZodValidator just fails.

Alikznollet: No need for this if because of this.

MrLegitFTW (approved):
This seems like a great idea to clean up the db services, only wondering if this removes part of the control we had beforehand? Though I suppose we don't realy use it or I can't think of an example at least.

Alikznollet:
> This seems like a great idea to clean up the db services, only wondering if this removes part of the control we had beforehand? Though I suppose we don't realy use it or I can't think of an example at least.

The control partly stays though, all I'm doing is making it easier (and more readable) to make generic clauses like the column names, or update fields. That way the objects can be reused to create these.

We can still add to the queries any way we want, just a little less duplicated code.

MrLegitFTW:
> > This seems like a great idea to clean up the db services, only wondering if this removes part of the control we had beforehand? Though I suppose we don't realy use it or I can't think of an example at least.
> 
> The control partly stays though, all I'm doing is making it easier (and more readable) to make generic clauses like the column names, or update fields. That way the objects can be reused to create these.
> 
> We can still add to the queries any way we want, just a little less duplicated code.

I'm talking about for instance you auto include all fields in the return clause of an object where if we for instance would like a default value or something then we couldn't under this: return Object.keys(schema.shape).join(", ");
However I believe we don't need that at the moment and is a minor price to pay for the advantages it brings, just thought I would bring it up.

Alikznollet:
> I'm talking about for instance you auto include all fields in the return clause of an object where if we for instance would like a default value or something then we couldn't under this: return Object.keys(schema.shape).join(", "); However I believe we don't need that at the moment and is a minor price to pay for the advantages it brings, just thought I would bring it up.

Okay so if I understand correctly you mean we wouldn't be able to for example return a default value together with the object? Or? Because technically you could add more fields after the return clause.

Alikznollet:
I've fixed the changes!

backend/src/database/db-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+ */
+
+/**
+ * Generates a COUNT query for the table provided. Looks at the id field.
+ * @param tableName The table name to count.
+ * @returns The query.
+ */
+export function generateCountQuery(tableName: string): string {
+  const query = `
+      SELECT COUNT(DISTINCT id) as count
```
MatthieuDC: Using DISTINCT here adds massive unnecessary overhead, since id is inherently unique

backend/src/database/db.blog.service.ts
```diff
@@ -53,61 +60,52 @@ export class BlogDatabaseService {
...
-    let i = 1;
+    const param = (val: any) => {
+      values.push(val);
+      return `$${values.length}`;
+    };
 
     // Title filter
     // NOTE: This is case-insensitive and looks in all languages + matches on parts.
     if (blogFilters.title) {
+      const titleParam = param(blogFilters.title);
```
MatthieuDC, line 75: Does this still match partials?

backend/src/database/db.price.service.ts
```diff
@@ -20,14 +27,20 @@ export class PriceDatabaseService {
...
+    const query = `
+      SELECT ${returningClause}
+      FROM prices 
+      WHERE id = $1;
+    `;
+
     const result = await this.db.query<PriceDto>(query, [id]);
 
+    if (result.length === 0) {
+      throw new ResourceGoneException(`Location with ID ${id} not found.`);
```
MatthieuDC: This should be Price instead of Location

MatthieuDC (changes requested):
Mostly looks very good! I only have a couple of remarks here, would love it if you could take a look. Rest looks amazing

backend/src/database/db.blog.service.ts
Alikznollet:
```ts
const titelClauses = SUPPORTED_LANGUAGES.map(
  (lang) => `titel->>'${lang}' ILIKE ${titleParam}`,
);
conditions.push(`(${titelClauses.join(" OR ")})`);
```

Yes it does using these.

MatthieuDC (approved):
All is fixed, thanks!

backend/src/database/db-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+    values,
+    nextIndex: index,
+  };
+}
+
+/**
+ * Filtering
+ */
+
+// In your db-utils.ts (or wherever you keep utilities)
```
landgeet: you forgot to remove 🤖 comment :)

backend/src/database/db-utils.ts
```diff
@@ -0,0 +1,179 @@
...
+      // If it's an array, use Postgres ANY() for an IN clause
+      if (Array.isArray(val)) {
+        conditions.push(`${pre}${col} = ANY(${param(val)})`);
+      } else {
+        // Standard exact match
+        conditions.push(`${pre}${col} = ${param(val)}`);
+      }
+    }
+  }
+}
```
landgeet, line 150: Can you add a comment here explaining this function and the params (like usual)?

backend/src/database/db.account.service.ts
```diff
@@ -90,13 +91,15 @@ export class AccountDatabaseService {
       WHERE username = $1
       LIMIT 1
     `;
-    const result = await this.db.query(query, [account.username]); // no type for safety.
+
+    // Returning AccountDto should be safe here since we never show this to the user.s
```
landgeet: small typo :)
```suggestion
    // Returning AccountDto should be safe here since we never show this to the users.
```

backend/src/media/services/media.crop.service.ts
```diff
@@ -57,44 +56,9 @@ export class MediaCropService {
    * @param autoDownload Whether to automatically fetch and save the physical file. Defaults to false.
```
landgeet: You forgot to delete this comment

landgeet (changes requested):
Looks really good!
I do have a question:
- Why did you remove the autodownload in the media?
Also some small nitpicky remarks below.

backend/src/database/db-utils.ts
Alikznollet: Oop

backend/src/database/db-utils.ts
Alikznollet: Will do, didn't realise I forgot this.

backend/src/media/services/media.crop.service.ts
Alikznollet: Will remove

Alikznollet:
> Looks really good! I do have a question:
> 
>     * Why did you remove the autodownload in the media?
>       Also some small nitpicky remarks below.

Autodownload isn't needed because the scraper is already automatically downloading media that isn't stored at our own adress. So no need for it.

landgeet:
> > Looks really good! I do have a question:
> > ```
> > * Why did you remove the autodownload in the media?
> >   Also some small nitpicky remarks below.
> > ```
> 
> Autodownload isn't needed because the scraper is already automatically downloading media that isn't stored at our own adress. So no need for it.

Makes sense thanks!

Alikznollet:
@landgeet Applied your changes!

MrLegitFTW (approved):
Looks thorough and fine by me.

landgeet (approved):
Ready to merge!

PR #336, opened 2026-04-09
Prints page

### LINKED ISSUE #332: Prints page

ddndauw:
## 🎯 MAIN GOAL
- [x] this PR implements a the prints page

## 🛠️ TESTING
There are no extra tests written for this code.

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Run `npm run start:dev`
2. Go to the prints page

## ⚠️ REMAINING ISSUES
\-

\-

## ✅ CLOSED ISSUES
- #332


Alikznollet (approved):
Visually looks stunning (love the cats!!!). Code wise I also can't find any problems. All of the functionality seems to work. Approved.

jillvanham (commented):
Looks great! The prints page is coming along nicely. I have a few small remarks:

- The filter translations are working correctly, but the section titles are not. For example: it shows "Affiche" in both English and Dutch.

- Because the filters are hidden by default, it wasn't immediately obvious that I was only looking at a subset of prints. Some suggestions:
-> Maybe the default should be 'all prints', and not 'poster'. Now you can never see all prints at once. 
-> If you prefer keeping 'affiches' as default, we should make it more obvious that a category is active. This could be done by keeping the filter bar visible by default or using a dropdown/select box instead of buttons.

- I love the new filters button! We should make this the default filter button and also use this on the archive page. 

- It might be nice to add a description field for each print, which could be displayed on the individual detail/ID page of a print, if that's something we want to implement. But this could be moved to a new issue.

ddndauw:
> * The filter translations are working correctly, but the section titles are not. For example: it shows "Affiche" in both English and Dutch.

I have fixed this !
> * Because the filters are hidden by default, it wasn't immediately obvious that I was only looking at a subset of prints. Some suggestions:
>   -> Maybe the default should be 'all prints', and not 'poster'. Now you can never see all prints at once.
>   -> If you prefer keeping 'affiches' as default, we should make it more obvious that a category is active. This could be done by keeping the filter bar visible by default or using a dropdown/select box instead of buttons.

I have added an "All"-section, can you check if everything still works ? 

> * I love the new filters button! We should make this the default filter button and also use this on the archive page.

I will ask this in the Discord chat.

> * It might be nice to add a description field for each print, which could be displayed on the individual detail/ID page of a print, if that's something we want to implement. But this could be moved to a new issue.

We will do this in another PR as discussed on Discord

jillvanham (approved):
Thanks for the updates! One last small thing: for the translation of 'all,' I wouldn't use 'allemaal'. I would suggest using 'alles'. Everything else looks great. I'll approve it now so you can merge as soon as this small change is made.

ddndauw:
> Thanks for the updates! One last small thing: for the translation of 'all,' I wouldn't use 'allemaal'. I would suggest using 'alles'. Everything else looks great. I'll approve it now so you can merge as soon as this small change is made.

I have changed it !

PR #343, opened 2026-04-10
Filters for the blogs page

Robindr145:
## 🎯 MAIN GOAL
- [x] this PR implements the filters for the blogs page
## 🛠️ TESTING
- none, will be done when the page is complete

**Missing tests:**
* all

## 📝 DOCUMENTATION

**Missing documentation:**

## 🔍 HOW TO TEST
1. Go to `/blogs` via `npm run start:dev` or with the preview and click on the filter-button

## ⚠️ REMAINING ISSUES
- none

## ✅ CLOSED ISSUES
- #issuenumber


ddndauw:
Visually wise it looks really good, the filters have a very clean look!
I have a few small remarks/suggestions still:
Robin1

- On smaller screens range, after and before look a bit weird, I think there is enough place to still put them horizontally?
- Is there a reason why end date and start date have a different layout? (I'm asking because the frame surrounding oldest has the same color as the background.)
- Right now the dropdown is selecting a year range I think? Maybe make it so a user can select 2 years max there to avoid confusion?

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,286 @@
...
+  It contains two sections:
+    1. Year multiselect  — derived from oldestDate/newestDate backend bounds.
+       Multiple years = union mapped to a single before/after range
+       (min-year Jan 1 → max-year Dec 31).
+    2. Calendar          — fine-grained date picking.
+  Using one clears the other.
+-->
+<script lang="ts" setup>
+import type { FilterBlog } from "@repo/common";
+import Calendar from "~/components/DefaultCalendar.vue";
```
ddndauw: You can use Nuxt's auto-import by removing this and using DefaultCalender below. But feel free to decide for yourself which option is best. :>

Robindr145:
@ddndauw well noted about that phone version. I had not tested that.
I have not changed the color of the "today" and "oldest" in this PR, but you noticed that well, I will also change it the oldest in purple. 

The dropdown only selects 1 year, as discussed in the Discord.

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,286 @@
...
-}
-</style>
+<!--
+  components/blogs/StoryToolbar.vue
+  ===================================
+  Toolbar row + in-flow filter panel.
+
+  The panel opens below the toolbar (in flow), pushing <main> down.
+  It contains two sections:
+    1. Year multiselect  — derived from oldestDate/newestDate backend bounds.
```
landgeet: i think you forgot to update this comment because this is now a single year select, not multiyear

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,286 @@
...
+  if (!props.oldestDate || !props.newestDate) return [];
+  const start = new Date(props.oldestDate).getFullYear();
+  const end = new Date(props.newestDate).getFullYear();
+  const arr: number[] = [];
+  for (let y = end; y >= start; y--) arr.push(y); // newest first
+  return arr;
+});
+
+function toggleYear(year: number) {
+  selectedYear.value = selectedYear.value === year ? null : year;
```
landgeet: this line will never get triggered because the radio thing doesn't call `change` when clicking the same year

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,286 @@
...
+const panelOpen = ref(false);
+const selectedYear = ref<number | null>(null);
+
+watch(searchQuery, (val) => emit("update:search", val));
+
+// Year range
+const years = computed<number[]>(() => {
+  if (!props.oldestDate || !props.newestDate) return [];
+  const start = new Date(props.oldestDate).getFullYear();
+  const end = new Date(props.newestDate).getFullYear();
```
landgeet: Does this work with different timezones? For example `new Date("2024-01-01").getFullYear()` can return 2023 in some timezones.

landgeet (changes requested):
Visually this looks great!
There is a functional issue i found though:
The Date tab doesn't seem to work when testing on dev.
E.g. there are two stories on april 1st and i tried to filter on it and got nothing (see picture below)

Image

Also when i select one or mutliple dates and then press a year it only filters on the year while the dates stay selected, maybe you should deselect the dates when filtering on year only? (see image fo clarity)

Image

There are also some remarks in the code below

Nowak-Dominik:
@Robindr145 
Great job in general!

However I did notice a few small things:

## Clear button
Maybe try to make it a little bit more noticible, it's not bad, but maybe it could be more visible somehow. I think that now it becomes visible kind of far away on the right side and becouse it is small to match the size of the "Year" label it is not too much visible. If you don't come up with a better way to display this button, then it is fine, becouse it is not the end of the world honnestly.

image

By the way maybe the location of that clear button is not the problem, maybe just add something which makes it stand out a little more. Maybe try an underline or something else.

## Calendar not updating visually
So I have noticed an issue that I had as wel when I used the calendar.
After you select something in the calendar and then use the years above to select a year, the calendar filters get cleared but not visually.

### Using calendar:
image

### Using the year selector after using the calendar:
image

As you can see the calendar clearly selects a range but the year 2022 overrides the filters without updating the calendar looks. I suggest checking out the way I fixed this on my archive PR it. If I remember correctly you must assign a key to the calendar and if the year is selected you can incrementa that key, so the UI is forced to re-render the view.

Nowak-Dominik:
@Robindr145 
I like the changes, thanks
I approve it now look wise
I do have to check the code still to approve it in general

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,341 @@
...
-  max-height: 600px;
-}
-.cal-slide-enter-from,
-.cal-slide-leave-to {
-  opacity: 0;
-  max-height: 0;
-}
-</style>
+<!--
+  components/blogs/StoryToolbar.vue
```
Nowak-Dominik, line 2: Maybe add some more documentation here, just shortly so it is clear what this is for.

frontend/app/components/blogs/StoryToolbar.vue
```diff
@@ -1,143 +1,341 @@
...
+            <select
+              v-model="sortOrder"
+              class="h-10 px-3 rounded bg-muted border border-border text-[10px] font-brand font-black uppercase tracking-widest text-muted-foreground focus:outline-none cursor-pointer transition-colors hover:border-foreground/30 w-max"
+            >
+              <option value="newest">{{ t("stories.sortNewest") }}</option>
+              <option value="oldest">{{ t("stories.sortOldest") }}</option>
+            </select>
+          </div>
+
+          <!-- Year picker -->
```
Nowak-Dominik, line 200: Maybe this year picker could be a component on it own defined in a different file

Nowak-Dominik (changes requested):
@Robindr145 
This PR looks good, but could you check the things I pointed out and see if you could try to do them.

Also I wil approve this after you resolve the merge conflict.

Nowak-Dominik (approved):
@Robindr145 
Great work, thanks for the changes!

Don't forget to resolve the merge conflict before merging

landgeet (approved):
This looks great now!
I like the new layout and everything seems to work properly.
Great work!
(As already said, don't forget to resolve the merge conflicts)

PR #348, opened 2026-04-12
Prints Enum

### LINKED ISSUE #347: Print Type Endpoints
Print Endpoints have to be changed to contain a field for the type of print

MatthieuDC:
## 🎯 MAIN GOAL
* implements different types of prints in the endpoints

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. The usual endpoints stuff
2. Check getPrints for filtering on types

## ✅ CLOSED ISSUES
- #347


common/src/objects/print_item.ts
```diff
@@ -5,12 +5,23 @@
...
 // Base Print Item object.
 export const PrintItemSchema = z.object({
   id: z.number(),
   titel: LocalizedStringSchema,
   description: LocalizedStringSchema,
   url: z.string(),
+  print_type: PrintTypeZodEnum,
   created_at: z.string().datetime(),
   updated_at: z.string().datetime(),
 });
```
landgeet: I would use the same structure used for the crop names in `media_crops.ts`, this avoids of using deprecated functions like `nativeEnum`. You could also export both a normal list of the types like suggested and additionally return an enum so you can use it elsewhere.
This is just a suggestion so you can do with it what you want.

backend/src/print/print_item.controller.ts
```diff
@@ -64,16 +66,26 @@ export class PrintItemController {
...
+    required: false,
+    description: "Filter by print type.",
+  })
   @ApiOkPaginatedResponseAnyOf(PrintItemDto, PrintItemViewDto)
   @Get()
   async getPrintItems(
     @Query(new ZodValidationPipe(PaginationFilterSchema))
     paginationFilter: PaginationFilterDto,
     @Query(new ZodValidationPipe(LanguageQuerySchema)) lang: LanguageQueryDto,
+    @Query("type") print_type?: PrintType,
```
landgeet: Maybe it would be nice to have a validationPipe here too. You could do this by making a seperate schema for the enumtype and putting a validationPipe around it.
I have no idea what is best practice in this case, so this could be the correct way to do it anyways.

landgeet (changes requested):
This looks really good, great work!
There are some remarks i have below.
These remarks are purely me not knowing if what you did is best practice or not.
So i would ask @Alikznollet to see if my comments are correct or not :)

common/src/objects/print_item.ts
Alikznollet: I second this, maybe have a look at how we managed this for things like CropName or GalleryType.

backend/src/print/print_item.controller.ts
Alikznollet: Same goes for here. Have a look at how Seb handled the GalleryType enum.

Alikznollet (approved):
This seems to work correctly! I approve :)

MrLegitFTW (approved):
This looks good.

backend/src/print/print_item.service.ts
```diff
@@ -6,8 +6,8 @@ import {
   PaginationFilterDto,
   PrintItemDto,
   ReplacePrintItemDto,
-  PrintType,
 } from "../dto/dto";
+import type { PrintType } from "../dto/dto";
```
landgeet, line 10: make sure to group these imports

landgeet (approved):
Looks good now! (maybe just fix the import grouping i mention below before merging)

PR #368, opened 2026-04-14
Fix Caching + New Backend Exceptions

### LINKED ISSUE #335: Fix Caching Issues
- [ ] Some GET requests are cached and will not change immediately when a resource is updated. (Blog media for example)
- [ ] Replace all 410 exceptions with 422 for this.

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* Fixed a bug with caching that would cause the backend to serve wrongful information to the user.
* To achieve this introduced three new NotFoundException subclasses that can be thrown in the right situations. All `ResourceGoneException` instances have been replaced with these more unified classes.
  * `ResourceNotFoundException`: When a Resource does not exist currently.
  * `LinkNotFoundException`: When you're unable to retrieve data from a link.
  * `MediaNotFoundException`: When a certain object has no media of the given type.
* Removed all error handling from delete functions since they should by REST principles be idempotent (doesn't matter if something is already deleted, it just has to be gone after the request.)

## 🛠️ TESTING
- [x] all tests succeed

Tests that needed altering already changed.

## 📝 DOCUMENTATION

Documentation has been updated accordingly.

## 🔍 HOW TO TEST

You can go through the API via Swagger and just test some endpoints see if they still work. Though this didn't change a lot.

You can see the caching work accordingly now by doing the following:

1. Look for a `Production` that has no Gallery attached. And ask for that via the endpoint. You will receive a `MediaNotFoundException.`
2. Then link any gallery to this production and query it again.
3. The gallery should show up properly.

## ✅ CLOSED ISSUES
- Closes  #335


landgeet (approved):
I like this change a lot!
This makes it so the user is way better informed with errors.
The code also looks clean, so no remarks from my end.

Nowak-Dominik (approved):
@Alikznollet 

Everything looks fine to me. I like this update becouse it makes the exceptions more clear, good work!

PR #373, opened 2026-04-14
Admin: login page

### LINKED ISSUE #358: Admin: login page

Nowak-Dominik:
…ductions

## 🎯 MAIN GOAL
* Login page for the admin website

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* All

## 📝 DOCUMENTATION

**Missing documentation:**
* None

## 🔍 HOW TO TEST
1. Go to `/admin/login`

## ⚠️ REMAINING ISSUES
Not quite a bug but something to think about in the future. After loging in and then going back to the main site, the header acts like we are stil loged in so it does not show the normal header but the admin header. This should be fixed in other PR becouse it is a Header issue. Maybe something like going back to the normal site should automaticly log out the user.

There are some screenshots posted underneath in a comment

## ✅ CLOSED ISSUES
- #358


Alikznollet:
@Nowak-Dominik Maybe a little early, but it might be nice to have the Login page not even have the footer or the header. So just only the login thing, and also to make it not scrollable.

Being able to scroll and the header are pretty distracting haha.

MrLegitFTW:
note: we will be turning this test off while debugging the issue further.

Nowak-Dominik:
> @Nowak-Dominik Maybe a little early, but it might be nice to have the Login page not even have the footer or the header. So just only the login thing, and also to make it not scrollable.
> 
> Being able to scroll and the header are pretty distracting haha.

@Alikznollet 

Yeah I was thinking about that as wel, good to see that you share my thoughts.
I have to find a way to somehow disable the standard template so that this login does not get the header and footer automaticly. I wil check it out today.

frontend/app/pages/admin/login/index.vue
```diff
@@ -0,0 +1,120 @@
...
+            <span v-if="loading">{{ t("login.signingIn") }}</span>
+            <span v-else>{{ t("login.signIn") }}</span>
+          </button>
+        </form>
+      </div>
+    </div>
+  </div>
+</template>
+
+<style scoped></style>
```
Alikznollet: Maybe remove this since no styles are being used. Very small remark.

Alikznollet (commented):
Code looks sublime. One small remark about an empty style tag and something I would like to suggest:

Currently because the header and footer are gone (which is very nice btw) you can't change from light to dark mode in this page. Maybe we could re-use that button and put it somewhere in the top right? But without the header then, just the button for changing between modes.

Also, maybe you could make `/admin` route to `/admin/login` if you're not logged in?

Nowak-Dominik:
> Code looks sublime. One small remark about an empty style tag and something I would like to suggest:
> 
> Currently because the header and footer are gone (which is very nice btw) you can't change from light to dark mode in this page. Maybe we could re-use that button and put it somewhere in the top right? But without the header then, just the button for changing between modes.
> 
> Also, maybe you could make `/admin` route to `/admin/login` if you're not logged in?

Yeah good point. I was also thinking about somehow introducing the light/dark and language buttons again without the header or something, becouse it is a shame that the user cant select a language nor a different theme. Especialy becouse the dark mode looks nice, at least I think it looks good.

I did not want to risk adding more code so the test don't fail somehow but maybe this is worth the risk.

And your second suggestion is also intresting, I wil think about it.

Robindr145:
I see in the screenshots that the "eye" icon to see your password is not present; could you add this?

Nowak-Dominik:
> I see in the screenshots that the "eye" icon to see your password is not present; could you add this?

Yes great point, I did not thought about this. Wil do!

Nowak-Dominik:
This is the updated look:

image

image

Nowak-Dominik:
> note: we will be turning this test off while debugging the issue further.

Note update:
We did not turn that test off, it somehow fixed itself.

Alikznollet (approved):
This looks excellent now!

frontend/app/components/ThemeToggle.vue
```diff
@@ -0,0 +1,42 @@
+<script setup lang="ts">
+import { ref, onMounted } from "vue";
+import { Sun, Moon } from "lucide-vue-next";
+
```
MrLegitFTW, line 10: perhaps add a description in this file of what the purpose is?

frontend/app/pages/admin/login/index.vue
```diff
@@ -0,0 +1,153 @@
...
+// This makes sure that the login does not use the default layout with the header and footer.
+definePageMeta({
+  layout: false,
+});
+
+const togglePassword = () => {
+  showPassword.value = !showPassword.value;
+};
+
+async function handleLogin() {
```
MrLegitFTW, line 42: perhaps a desc for this function too? It is definitly clear as is, so it is up to you.

MrLegitFTW (approved):
I have small remarks regarding readability, seeing as these are entirely optional I will approve this. Great job on this.

MrLegitFTW:
+ extra question: Could you look into making sure you cant go to any admin pages without having logged in first? Perhaps is a bit out-of-scope for this pr but is worth looking into imo.

Nowak-Dominik:
> * extra question: Could you look into making sure you cant go to any admin pages without having logged in first? Perhaps is a bit out-of-scope for this pr but is worth looking into imo.

Just to be clear, as I mentioned on Discord, I already looked into this and spent quite a bit of time on it, but it didn’t work. So I created a separate issue for it, and it will be resolved in a different PR.

PR #386, opened 2026-04-16
Fix Print Item Filter + Blog Filter

### LINKED ISSUE #346: [FIX] check before-blog filter if it works as intended
The before ... date filter for the blogs is an exclusive filter instead of an inclusive filter (I think). 
If this is the case, please fix the code in `frontend/app/components/DefaultCalendar.vue`:

```ts
const filter = computed<DateFilter>(() => {
  switch (mode.value) {
    case "single":
      return selected.value
        ? { after: selected.value, before: addDays(selected.value, 1) }
        : {};

    case "range": {
      if (!anchor.value || !selected.value) return {};
      const [a, b] = sorted(anchor.value, selected.value);
      return { after: a, before: addDays(b, 1) };
    }

    case "after-selected":
      return anchor.value
        ? { after: anchor.value, before: addDays(todayIso.value, 1) }
        : {};

    case "before-selected":
      if (!props.oldestDate) return {};
      return selected.value
        ? { after: props.oldestDate, before: addDays(selected.value, 1) }
        : { after: props.oldestDate };
  }
});
```

And remove the addDays-function in `/frontend/app/utils/formatters.ts`

### LINKED ISSUE #387: [FIX] Add filter by name to prints

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* Print items now have filters for their `getAll` endpoint. You can ask for a `title` filter that will try to match with any language partially and the `type` is now also in the filter object.

* Fixed a bug where the `Blog` filter for `title` wouldn't match with anything partially.

* Fixed a bug where the `Blog` date filters weren't inclusive. They now are and the frontend has been adjusted according to the issue description. 

## 🛠️ TESTING
- [x] all tests succeed

The tests have been altered for the new values.

## 📝 DOCUMENTATION

No changes here.

## ✅ CLOSED ISSUES

- Closes #346 
- Closes #387

### DIFF
--- a/backend/src/database/db.print_item.service.ts
+++ b/backend/src/database/db.print_item.service.ts
@@ -6,7 +6,7 @@ import {
   ModifyPrintItemDto,
   PaginationFilterDto,
   ReplacePrintItemDto,
-  PrintType,
+  FilterPrintItemDto,
 } from "../dto/dto";
 import { PaginatedResponse } from "@repo/common/src/objects/pagination";
 import {
@@ -14,7 +14,7 @@ import {
   generateReturningClause,
   generateUpdateClause,
 } from "./db-utils";
-import { PrintItemSchema } from "@repo/common";
+import { PrintItemSchema, SUPPORTED_LANGUAGES } from "@repo/common";
 import { ResourceNotFoundException } from "../common/exceptions";
 
 @Injectable()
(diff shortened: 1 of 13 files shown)


Alikznollet:
@MrLegitFTW @landgeet @Robindr145 @ddndauw I think the four of you are the most familiar/benefit the most from these changes. This PR is now done and would need to be merged so @ddndauw can continue work on #336.

ddndauw (approved):
I don't see anything fishy here. 👍Thank you for implementing/updating the filters.

backend/src/database/db.print_item.service.ts
```diff
@@ -52,33 +52,62 @@ export class PrintItemDatabaseService {
...
+    // Filter for prints of a certain type
+    if (printItemFilters.type) {
+      conditions.push(`print_type = ${param(printItemFilters.type)}`);
+    }
+
+    const whereClause = conditions.length
+      ? `WHERE ${conditions.join(" AND ")}`
+      : "";
+
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
MrLegitFTW, line 90: I feel like this could be fixed instead of disabling it?

MrLegitFTW (changes requested):
I don't think we should encourage the disabling of the eslint warnings and feel like this could easily be fixed instead of being turned off here. Besides that this looks great.

backend/src/database/db.print_item.service.ts
Alikznollet: I'll look into how to fix this but I'm pretty sure it's almost impossible. Because the values in that array are inherently any.

MrLegitFTW: Just looked into it a bit more and it would indeed be a bit too much of a pain to fix, saw we do the same in pretty much all other places in the db services, so for readability sake I would suggest to just keep as is.

MrLegitFTW: Could perhaps make this an issue for milestone 3 if we really want to.

Alikznollet: It's fixed in my next PR that is leaning on this one.

MrLegitFTW (approved):
Looks great. Can be merged. Will talk about solving the eslints later (if we even want to).

Issue #385, opened 2026-04-16
Accounts page: Create account panel component


landgeet:
- create a component for creating an account

PR #388, opened 2026-04-17
Account panel

landgeet:
## 🎯 MAIN GOAL
* Add an AccountForm implementation of the BaseForm component.
* Add password input option with eye to show/hide password.
* Change BaseInput so letters are not all uppercase.
* I currently commented the superadmin part if we want to add multiple superadmins.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Run `npm run start:dev`
2. Go to `http://localhost:3001/tempShowcase`

Make sure to checkout the light and dark mode and different screen sizes.
**Disclaimer:** the dark mode is the same as for the BaseForm so if we want to change it, we change the colors for BaseForm.
Also check mobile version if the list looks good.

## ✅ CLOSED ISSUES
- #385


frontend/app/components/admin/AccountForm.vue
```diff
@@ -0,0 +1,96 @@
...
+    component: "BaseInput",
+    name: "password",
+    props: {
+      label: t("accounts.password"),
+      type: "password",
+      placeholder: t("accounts.password_placeholder"),
+      required: true,
+    },
+  },
+  /*
```
MrLegitFTW, line 90: Perhaps add some explenation as to why this is here and commented out. (a bit about allowing the possibility of multiple accounts?)

MrLegitFTW (approved):
Code-wise this looks great. Can't really say anything. I have a small suggestion for a bit of explenation where I think it could be beneficial. However I leave that up to you. Stylistically I can't really say much as that is not my department. Great job on this!

frontend/app/components/admin/AccountForm.vue
landgeet: i added a small comment now

ddndauw:
I have not much to say, this is okay to be merged imo ! 

- I think the layout is a bit outdated compared to the rest of the site, if you want to, you can try to make it match the site a bit more (by adding a bit more of the purple color for example.) But if that is not your cup of tea, we can leave this for another PR too.  
- Some sites have two passwords fields for creating accounts, (like entering the same password twice and then checking if they're the same). I don't know if that is useful here or a bit too "extra".
- Maybe you can make BaseInput accept a minimum character count? Right now you can type in a username and password with one character and it feels a bit illegal..

Other than these I don't see/think of anything else, the panel looks good on smaller screens too !
Code-wise I don't have any remarks.
Let me know what you think of these suggestions? If you feel like what you have now is alright already I'll approve !

landgeet:
@ddndauw 
I agree on all of your suggestions!
I applied the latter 2 but I will leave the layouting to someone who is more familiar with it for a future pull request.
Let me know what you think!

Alikznollet (approved):
This looks ready to be merged!

ddndauw (approved):
> I agree on all of your suggestions!
> I applied the latter 2 but I will leave the layouting to someone who is more familiar with it for a future pull request.
> Let me know what you think!

Thank you for the changes! I'll approve of the PR!
Maybe make a separate issue for the layout ? (Just in case we forget haha, like a general issue "improve forms layout", if you want to I can make the issue too instead.)

PR #391, opened 2026-04-17
Refine Existing + Add New Filters

### LINKED ISSUE #334: Refine Filters.
- [x] * Production Text Filters should look at titel and artist.
- [x] * Look for tags by their title in getAll
- [x] * Look for locations by their name.
- [x] * Fix production date filters to be more intuitive. We don't have to filter for time that's useless. (See BLOGS for an example.)

Alikznollet:
## 🎯 MAIN GOAL

* Implemented filters for both Locations and tags that can be used in the admin part.
* Reworked the date filters for productions.
* Reworked the title and artist filter for productions.

* The productions page only shows finished productions now!!

* Fixed a discrepency with the blog date filters (they're faster now)
* Removed any warnings from the db services.

## 🛠️ TESTING
- [x] all tests succeed

* Tests shouldn't really change much.

>[!NOTE]
>Might have to change some of the tests where new filters were introduced.

## 📝 DOCUMENTATION

* Documentation inline has been updated.

## 🔍 HOW TO TEST

Go to swagger (or the frontend if the filters are already used in there) and try them all out.

## ✅ CLOSED ISSUES

- Closes #334

### DIFF
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -66,8 +66,8 @@ export class BlogDatabaseService {
     const returningClause = generateReturningClause(BlogSchema);
 
     const conditions: string[] = [];
-    const values: any[] = [];
-    const param = (val: any) => {
+    const values: (string | number)[] = [];
+    const param = (val: string | number) => {
       values.push(val);
       return `$${values.length}`;
     };
@@ -99,7 +99,6 @@ export class BlogDatabaseService {
       ? `WHERE ${conditions.join(" AND ")}`
       : "";
 
-    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
     const filterValues = [...values];
     const countQuery = `
       SELECT COUNT(*) as count FROM blogs
--- a/backend/src/database/db.blog.service.ts
+++ b/backend/src/database/db.blog.service.ts
@@ -84,12 +84,15 @@ export class BlogDatabaseService {
 
     // Filter blogs that were created before.
     if (blogFilters.before) {
-      conditions.push(`created_at::date <= ${param(blogFilters.before)}::date`);
+      conditions.push(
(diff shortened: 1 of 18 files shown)


common/src/objects/productions.ts
```diff
@@ -44,8 +44,7 @@ export const ReplaceProductionSchema = MutableProductionSchema;
 
 // Filtering.
 export const FilterProductionSchema = z.object({
-  titel: z.string().optional(),
-  id: z.coerce.number().optional(),
```
landgeet, line 48: Is there a reason you removed Id here, did we just never use it in the filters?

landgeet (approved):
Looks good and everything seems to work!
Just a small question below for clarity.

backend/src/database/db.blog.service.ts
```diff
@@ -84,19 +84,21 @@ export class BlogDatabaseService {
 
     // Filter blogs that were created before.
     if (blogFilters.before) {
-      conditions.push(`created_at::date <= ${param(blogFilters.before)}::date`);
+      conditions.push(
+        // NOTE: We add + 1 day here for performance and include reasons.
```
MrLegitFTW, line 88: perhaps add here that the performance reason is due to indexing in the db? That way  it is more clear.

MrLegitFTW (approved):
Looks great, if you want I would suggest a small comment you could add for clarity. Can be merged.

common/src/objects/productions.ts
Alikznollet: Yeah there's actually a lot of filters we don't use in there right now, but i left most of them in here. The id filter is gone because why use it when we could just get the production by id.

backend/src/database/db.blog.service.ts
Alikznollet: Sure thing.

PR #403, opened 2026-04-20
Admin accounts page (super-admin)

### LINKED ISSUE #362: Admin: accounts page

landgeet:
## 🎯 MAIN GOAL
* Implements the accounts page for creating/deleting accounts by super admins
* Adds and uses the AccountAlreadyExists error (written by @MrLegitFTW )
* Defined a reset ref so the fields reset after submitting an acocunt
* Add snackbar for error/succes messages

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Run `npm run start:dev` in project root. (or checkout the preview if production is fixed)
2. Navigate to `http://localhost:3001/admin/login` and log in with a super admin account.
3. Navigate to `http://localhost:3001/admin/accounts` while loged in.
4. Check the accounts page and try things out like creating and deleting accounts.

Make sure to check that the accounts page is not accesible through a normal admin account (you can make one yourself to check this)

## ✅ CLOSED ISSUES
- Closes #362


Alikznollet (approved):
Reviewed both code and visuals. Looks absolutely amazing!!

Nowak-Dominik:
@landgeet 

This looks great man, good job!

However I have a few things to point out:

## Confirmation

Now after creating or removing an account there is a nice green confirmation:

image

However, this stays there forever or untill I refresh. Try to work with a confirmation that shows only for a few seconds and then disappears.

## Misleading Dutch label for reset button

I think that it would be better if you would use a different word than "herstellen" for a button that clears all input fields.

image

When I saw this at first I was confused to what this button was for. First I thought it would recover a recently removed account or something like that. The english "reset" that you used feels better and is fine. But maybe it would also be better if you use something like "clear" instead of "reset"?, that I leave up to you. For the Dutch one you can use something like "legen" or "leegmaken".

## Labels for submit button

I think that it would be better to just call it "Create" and "Aanmaken" instead of "Submit" and "indienen".
But this one is fine as it is as wel, but I just think that it maybe would feel more fitting with the one I suggested.

All other things look great and are fine as they are to be approved.
I will check your code later today after the changes I just mentioned above are made.

Nowak-Dominik (approved):
@landgeet 

I checked the code and this looks good to me, I approve to merge it!

landgeet:
Currently the snackbar doesn't work on production, but #414 will fix this when looked at.

PR #404, opened 2026-04-21
Fix Verbose Console Prints + More

### LINKED ISSUE #400: [BUG] Stop printing errors in production console (vulnerability)
- This could give attackers the opportunity to use it to attack the website. We want to catch them in the frontend and not print them!

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* 404 Warnings were being printed out to the console when they aren't critical. This is now suppressed with a new onError function for those requests.
* The `tag_ids` filter for Productions was for some reason not optional. Is now fixed!
* The `Highlights` on the home screen were not filtering away the future productions. This now shows the same view as the productions page.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

Inline docs are there.

## 🔍 HOW TO TEST

Go to the site and open F12. Look at the in browser console and browse to a page that has media but no galleries.

To see the Highlights thing just go to the home page and see if the dates are correct.

## ✅ CLOSED ISSUES
- Closes #400

### DIFF
--- a/backend/src/auth/auth.spec.ts
+++ b/backend/src/auth/auth.spec.ts
@@ -48,7 +48,7 @@ describe("AuthGuards tests", () => {
       await request(app.getHttpServer() as Server)
         .get("/events")
         .expect((res) => {
-          expect([200, 410]).toContain(res.status);
+          expect([200, 404]).toContain(res.status);
         });
     });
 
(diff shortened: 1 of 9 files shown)


MrLegitFTW (approved):
Perfect, all is good. It is important to handle no media in code from now on without throwing the 404's seeing as this is very expected behavior.

landgeet (approved):
Everything seems to work now and the code looks good! No remarks from my end.

PR #407, opened 2026-04-21
Search Suggestions

### LINKED ISSUE #406: Search Suggestions
Implement backend level search suggestions for anything with a search bar.

Also apply them in the frontend.

Alikznollet:
## 🎯 MAIN GOAL

### Changelog

* Altered `SearchBar.vue` to take a function that returns `SearchSuggestion` objects that can be fetched from the backend.
* `SearchSuggestion` contains a `display` string (what's shown), a `context` string (optional secondary text) and `searchValue` which is the string that's actually searched if clicked.
* Added a new db-util function that generates a clause for the ordering that scores the rows. (used when suggestion searching)
* Added a psql extension that allows for way faster searches.
* Made the `before` date filter ACTUALLY work correctly...
* Cleaned up the filters for blogs view.

* Implementation in different pages:
  * Implemented this for the archive page. Uses an extra `is_suggestion` field in the productionFilters.
  * Implemented this for the blog paghe. Uses an extra `is_suggestion` field in the blogFilters.

> [!WARNING]
> Currently the active filters will also be applied to the search, which means you will only see suggestions that match the active filters. If you want this differently please do say so.

* TODO:
  * Update documentation.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

1. Run `npm i` in the root
2. Run `npm run start:dev` in the root
3. Go to the archive page on `https://localhost:3001/productions`. Start looking up some productions and see the suggestions updating.

Orrrr just test it on the preview.

## 📸 SCREENSHOTS:

afbeelding

## ✅ CLOSED ISSUES
- Closes #406


Robindr145 (changes requested):
This looks really nice to have, but I had an issue with it:
When I clicked on a suggestion, it showed the correct data, but when I removed the suggestion (with backspace instead of the cross), the filter was still "applied", it didn't reset to all the values anymore.

Image

Alikznollet:
> This looks really nice to have, but I had an issue with it: When I clicked on a suggestion, it showed the correct data, but when I removed the suggestion (with backspace instead of the cross), the filter was still "applied", it didn't reset to all the values anymore.

I mean the only thing I can do to fix that is to make sure the filter gets removed when the field is empty. But doesn't it make sense that the filter doesn't reset if you don't click enter?

ddndauw (approved):
Thank you for implementing search suggestions! Visually wise, I don't immediately see any issues. I also skimmed a bit through the code, and found nothing unordinary.
Active filters applied to the search is fine to me.
Also on Robin's comment, I don't see any issue with the filter still being applied when not clicking enter. When you are searching, the results also don't update until you press enter anyway? Unless I am understanding the issue wrong. I am approving this for now.

Robindr145 (approved):
Thanks for the changes !! This looks awesome.

PR #438, opened 2026-04-27
Prints description

### LINKED ISSUE #395: [Prints page] add description to grid item
The client requested a description to the grid item, we will further discuss how to do this, and if a detail page is needed.

ddndauw:
## 🎯 MAIN GOAL
- [x] this PR shows print description on grid view
- [x] this PR adds an edit page for the prints
- [x] this PR fixed images stretching

Extra:
- BaseForm now accepts an optional label for the submit- and resetbutton.
- Fixed "flashing" of base form components.

## 🛠️ TESTING
There are no tests written.

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Go to the prints page and look at the description of prints.
2. Edit a print on the prints admin page.

## ⚠️ REMAINING ISSUES
\-

\-

## ✅ CLOSED ISSUES
- #395


Alikznollet (commented):
Code wise this looks great! I would maybe like the description box itself a bit more centered when you click  the button, because right now it kind of sticks towards the top a bit. For the long one it also just feels like it's outside of the screen on the bottom and away from the top.

MrLegitFTW (approved):
Looks perfect code-wise. I have nothing to add. Would like someone else to look at this visually however before merging.

landgeet (commented):
Visually this looks absolutely amazing. I quickly skimmed over the code and this also looked great!
The only thing i would maybe add for user-friendliness is to be able to close the info popup by clicking outside of the box. But if you don't think this is necessary, i approve.

Nowak-Dominik:
@ddndauw 
I checked it out visually and it looks really nice.
However could you maybe change the look of this button now as wel so it does not have to be done late?

image

As you can see this arrow does not look well there, its to small and the button feels too bulky probably becouse of that.

frontend/app/components/form/fields/BaseFileUpload.vue
```diff
@@ -8,6 +8,10 @@
  *  - Selected filenames displayed below the input
  *  - Binds File objects via v-model
  *
+ *  In contrast to other base components, this component does not handle requirement check itself !!
```
Nowak-Dominik, line 11: Can you explain why it does not checkt the requirement here?

frontend/app/pages/admin/prints/edit/[id].vue
```diff
@@ -0,0 +1,232 @@
+<!--
```
Comment (Nowak-Dominik) [inline frontend/app/pages/admin/prints/edit/[id].vue:1]: I wonder if you maybe could actually merge the edit and create to one compoennt that has like a prop "edit" or "create" so the edit page is then just that component with prop "edit" and create is just that page with prop "create". I suggest this becouse I think that they share a lot of logic right?

Nowak-Dominik (commented):
@ddndauw 

Generally this looks good to be approved but could you check my suggestions first and see if you can apply them?

Nowak-Dominik:
@ddndauw 

I just noticed a few more things look wise.

So maybe you could make that X button a little bit more visible here:

image

And also another suggestion, maybe you could make it so of the user presses somewhere on the screen outside of that popup then that action should close that popup as wel.

Another think I found out is that I can scroll through the page if my mouse is on the popup but I think that this is counter intuitive no?

ddndauw:
@Alikznollet Can you check the description placement? I think I have fixed it, but I don't know how it looks on other screens. 

@landgeet Pop-up can now be closed by clicking outside of it !

@Nowak-Dominik 
- I have changed the edit page for prints to look more like the stories edit page, so the back button changed with it too.
- For the required indicator, I encountered some problems with the edit page when there is already an image existing, I had troubling passing it to the base component. 
The browser checks on the <input ... /> part of the base component, the part that handles file user input. I don't know how to manually let the browser know there is already a file in edit. So if I indicate the field as required, the browser performs a check and fails, even though there is already a file in edit. I need to indicate it as required for the required red star. 
I think this is fixable with manual checks maybe, instead of letting the browser perform checks, but I think this is a hassle to do and not for this PR. Do let me know if it's easy fixable I might have not have thought of other stuff.
I did Ctrl+Shift+F through the files really quick and saw that nobody else was using the component yet, so I thought it would be OK to change.
- The close button is a little bit darker, do I still need to make it bigger or anything else? You can also now click outside the pop-up to close it, and the page doesn't scroll anymore.
- I have made a shared form for both create and edit.

Nowak-Dominik (approved):
@ddndauw 

Thanks for the changes, everything looks good now, code wise and visual wise. Great job!

landgeet:
Thanks for the changes, really well done!

There is however a small bug i found, but i don't know if this has to do with my browser or not.
When i open the info of a print, the scrollbar of the normal page disappears so the page appear to be shifting a few pixels to the right when i open the info.

To recreate you can just open and close the info fast and i think you'll see it.
(if the browser matters, i use chrome on windows 11)

image

image

Alikznollet (approved):
You've implemented each of the features I wrote down really well! Thanks for updating it, especcially the placement of the box :)

Approved!

landgeet (approved):
good to go now

PR #444, opened 2026-04-28
Admin productions: create and edit page

### LINKED ISSUE #439: Admin productions: create and edit page

Nowak-Dominik:
## 🎯 MAIN GOAL
- [x] Create separrate logic for the production form
- [x] Create the visuals for the production form

## 🛠️ TESTING
- [ ] all tests succeed

**Missing tests:**
* All

## 📝 DOCUMENTATION

**Missing documentation:**
* TODO

## 🔍 HOW TO TEST
Go to admin productions and check the create and edit page out.

## ⚠️ REMAINING ISSUES
* None
* None

## ✅ CLOSED ISSUES
- #439


Alikznollet (approved):
Amazed by how well structured this code is, at first glance I don't see any problems. Visually it also looks stunning!

landgeet:
I've skimmed over the visuals really quickly and have some remarks.
First of all i want to say that this looks amazing, really good job.

Edit page:
- The edit page on the preview doesn't seem to work, i keep getting 500 server errors because of the series not working yet. I will review this when it is fixed by @MrLegitFTW .

Create page:
- Maybe it would be nice to have a check that checks if all necessary fields have been filled in when going to the next step. For example when going from "Core info" to "Tags" you could check if the title and description 1 have been filled in instead of at the end of the whole production.
- I like how simple the tag create is, but is there a way to add a translated version for the tag?
- There is a bug with the timezones in the event create where the displayed time is 1 hour later then the inserted hour (see picture)
image

Sometimes even 2 hours:
image

- I found a different strange bug on the event create. When you create an event but leave all fields empty, go to the next page, then go back, the event is gone. But if you know try to add a new event, it adds 2 instead of 1 (and it keep adding one if you repeat this process).
Add empty event:
image
Go to next and back:
image
Add new event:
image
- For adding media, it would be nice if next to the names (HD ready, HD square, etc) you could add text where it is used between brackets. This is already done in the blogs so you could look there for an example.
- It also would be nice if you would use the snackbar for the error handling do the user also sees what went wrong (like the "a production needs at least 1 event" is currently only shown in the console.)

Also just a general question from my end:
Where would you like me to add the csv uploads for the productions/events/tags/prices? Do i just put them underneath the component?

I know this is a lot of remarks, so i hope you can forgive me for nitpicking a bit 😅
If there are some changes/remarks you don't agree with, you can comment on it!

In general great job with this pr, i hope these changes will make it even better!

Nowak-Dominik:
> I've skimmed over the visuals really quickly and have some remarks. First of all i want to say that this looks amazing, really good job.
> 
> Edit page:
> 
> * The edit page on the preview doesn't seem to work, i keep getting 500 server errors because of the series not working yet. I will review this when it is fixed by @MrLegitFTW .
> 
> Create page:
> 
> * Maybe it would be nice to have a check that checks if all necessary fields have been filled in when going to the next step. For example when going from "Core info" to "Tags" you could check if the title and description 1 have been filled in instead of at the end of the whole production.
> * I like how simple the tag create is, but is there a way to add a translated version for the tag?
> * There is a bug with the timezones in the event create where the displayed time is 1 hour later then the inserted hour (see picture)
> 
> image
> Sometimes even 2 hours: image
> 
> * I found a different strange bug on the event create. When you create an event but leave all fields empty, go to the next page, then go back, the event is gone. But if you know try to add a new event, it adds 2 instead of 1 (and it keep adding one if you repeat this process).
>   Add empty event:
> 
> image Go to next and back: image Add new event: image - For adding media, it would be nice if next to the names (HD ready, HD square, etc) you could add text where it is used between brackets. This is already done in the blogs so you could look there for an example. - It also would be nice if you would use the snackbar for the error handling do the user also sees what went wrong (like the "a production needs at least 1 event" is currently only shown in the console.)
> Also just a general question from my end: Where would you like me to add the csv uploads for the productions/events/tags/prices? Do i just put them underneath the component?
> 
> I know this is a lot of remarks, so i hope you can forgive me for nitpicking a bit 😅 If there are some changes/remarks you don't agree with, you can comment on it!
> 
> In general great job with this pr, i hope these changes will make it even better!

I thank you for the time you took and These issues wil be fixed in another PR. It is best e don't wait with merging this one so we can start working further.

About some issues you mentioned:
* As you said yourself there is a error 500 becouse the series table is not yet added on prod, however it works on dev.
* It is a good idea to add like a requirenments list at the beginning / end so the user knows it but I don't want to do his each step becouse I want to give the user the freedom to jump between steps event before meeting the requirements.
* Good point, there is no way to add the translation tag, this is not that big of a problem but you are right, we could add that in the future.
* that bug about the time displaying later sometimes is crazy spotted, great job. I was not aware.
* I was not aware this events bug existed. I think I had it earlier but I tought that I fixed it, apparently not.
* The extra info remark for media in between brackets is a good point.
* Snackbar is a good point and that would fix the showing requirenments a little as wel.
* Maybe the best place for the CSV uploads would just be a separate button for that on the `admin/productions`.

PR #460, opened 2026-05-03
Error codes

### LINKED ISSUE #447: Backend errors
- add more variation to the backend errors.
- for example use more postgres error using the interface so we have less "internal server errors".

MrLegitFTW:
## 🎯 MAIN GOAL
* adds better error handling to the backend in general.
* adds missing JDOC documentation here and there.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. run npm run start:dev and use the backend like normal
2. OR see preview below.

## ⚠️ REMAINING ISSUES
* could possibly impact error handling elsewhere due to the new errors however all should be backwards compatible but you should check just to be safe.

## ✅ CLOSED ISSUES
- #447


backend/src/database/db.price.service.ts
```diff
@@ -24,7 +27,8 @@ export class PriceDatabaseService {
   /**
    * Get a single price by their ID.
    * @param id The ID we're trying to fetch.
-   * @returns The Price if there is one.
+   * @returns The Price if there is one.µ
```
Alikznollet:
```suggestion
   * @returns The Price if there is one.
```

Alikznollet (approved):
Looks good to me! One small comment I noticed with a typo.

MatthieuDC (approved):
Looks good! If you could change Alex's comment on the typo, everything is in order!

PR #467, opened 2026-05-07
add error page

### LINKED ISSUE #465: 404 Page
Create a 404 page for the website, be CREATIVE!

MrLegitFTW:
## 🎯 MAIN GOAL
* main error page for fatal errors
* fixes weird/broken page rendering in cases of bad network etc...

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. go to any page (that does not exist) to test the 404 error
2. load npm run:dev in /frontend to trigger the network errors.
3. surf to /test-500 and click on the button to trigger a 500 error

## ⚠️ REMAINING ISSUES
- [] there are still remaining bugs in the code
* Possibly added error handling in pages if they have custom handling internally. They will need to show fatal errors.

## ✅ CLOSED ISSUES
- #465


frontend/app/pages/productions/[id].vue
Comment (MrLegitFTW) [inline frontend/app/pages/productions/[id].vue:1]: note: the extra weird stuff is just my IDE being an ass, those can be ignored.

MrLegitFTW:
note: there is still a bug when going to a detail page through the url with all numbers like `productions/999999` you get a blank page instead of the error page. This is because nuxt seems to eat up all the errors I throw when this should be handled. No idea why. If anyone has any suggestions on fixing, all is welcome.

MrLegitFTW:
note to self: remove test page before merge.

frontend/app/error.vue
```diff
@@ -0,0 +1,96 @@
...
+      <!-- go home button -->
+      <button
+        class="px-8 py-4 bg-black text-white dark:bg-white dark:text-black rounded-full font-semibold text-lg hover:-translate-y-1 hover:shadow-lg transition-all duration-300"
+        @click="handleGoHome"
+      >
+        {{ t("error.goHome") }}
+      </button>
+    </div>
+  </NuxtLayout>
+</template>
```
Nowak-Dominik, line 94: Try to use the colours defined in the `tailwind.css` instead of having a hardcoded ones. Now you have like light and dark mode defined instead of just using an existing colour from tailwind so that this happends automaticly.

So try to use stuff like:
- `background `for background
- `foreground `for the text that is on the background
- `muted-foreground` for like slightly lighter text on the background

Nowak-Dominik (changes requested):
@MrLegitFTW 
PR looks good but I do want you to change the `class="..."` so that it uses the `tailwind.css` colours. 

After that it is ready to be approved from my side

frontend/app/error.vue
MrLegitFTW: will do.

Nowak-Dominik (approved):
@MrLegitFTW 
Great job! thanks for the changes

ddndauw (approved):
I have nothing to note here, everything looks good ! 

As for the bug that gives a blank page for non existing detail pages (with numbers), I don't know if it's fixed yet, but it still gives a blank page here, so I have looked/messed around for a bit, and haven't really found a solution myself. I tried to manually navigate to the error page when the response doesn't contain any data, which also resulted in a blank page. I tried more stuff but the page started tweaking so I give up and approve of this for now.

Image

If no one knows a solution maybe we can add the error layout in the template of the detail page itself ? But that is not very clean indeed.

MrLegitFTW:
new issue is created for that bug. In the meantime this will be merged. (#484)

PR #468, opened 2026-05-07
Redesign Homepage

### LINKED ISSUE #423: Home Page: Redesign
This is an EPIC issue that contains all issues regarding suggetions on the home page.

### LINKED ISSUE #449: Remove all informational content
Remove things like "about the archive" and other things from the home page.

### LINKED ISSUE #450: Home Page Search Component
Add a search bar and some quick links under it for years and tags to emphasize quick searches.

Alikznollet:
## 🎯 MAIN GOAL
* Home page redesign and some tweaks. 

Please don't hesitate to give visual feedback. But since I'm not really that good at the visual side of this project anyone is allowed to make or suggest changes!!

## 🔍 HOW TO TEST

Go to the preview and look at the homescreen. Test the search functionality and all of the filters.

## ✅ CLOSED ISSUES
- Closes #423 
- Closes #449 
- Closes #450


landgeet (commented):
I've looked at this visually and this looks absolutely amazing, well done!
I did find some small 'issues' which are just preferences.

Sometimes the tags carrousel just seems to end (see first picture)? After a couple of seconds the bar completes again but glitches a bit. To recreate it, just look at the bar for a minute or so and you'll see it. I don't know if that was intentional or not but i think it would be nicer if it kept going.

Carrousel ending on the right:
Image

Recovering but glitching a bit:
Image

There is also a small bug when you search something in the search bar, then go back to the homescreen and then try going back ot the searched page using the arrows, it doesn't save the search.
This does happens when searching and clicking on the page itself. This would be a small quality of life change if you could fix this.

Search something:
Image

Go back to the homepage (or a different page) and then go back using the arrows:
Image

landgeet:
Also the invalid end/start times for productions are messing up the search but we know this already.

frontend/app/composables/media/usePrintView.ts
```diff
@@ -6,7 +6,7 @@ import { ref } from "vue";
 
 import type { PrintItemView, PrintType } from "@repo/common";
 import { usePrintApi } from "./usePrintApi";
-import type { SearchSuggestion } from "~/components/SearchBar.vue";
+import type { SearchSuggestion } from "~/types/Search";
```
landgeet, line 9: I think there are some other files that still have this import unchanged (useBlogview for example) so you might want to check the other files too.

Alikznollet:
@landgeet 

First of all the thing about the tags is weird, for me it seems to happen only "sometimes". I'll look into a way to fix it but I'm not really sure how to at the moment.

The issue with the search is a bit annoying. Currently I'm wiping all search parameters when the user goes to the homepage, but that causes them to be wiped every single time you go to the homepage even if you press the back arrow.

What I might be able to do is only wipe the filters if the user starts typing in the search bar? Let me know what you think.

landgeet (commented):
The code looks splendid!
I really like how you did the different tiers for the suggestion search.
All the other code is also really well structured.
Some small remarks on the code though:
- Maybe you should add some tests for the more technical stuff (searchbar, useHomeView, etc)
- Add some comments at the top like in the other component/page files so the reader instantly knows what the component is used for and how it is used. The makes the code itself more user friendly

Other than that, the code quality is really high!

landgeet:
> The issue with the search is a bit annoying. Currently I'm wiping all search parameters when the user goes to the homepage, but that causes them to be wiped every single time you go to the homepage even if you press the back arrow.
> 
> What I might be able to do is only wipe the filters if the user starts typing in the search bar? Let me know what you think.

Yea this is a tough problem. Is there a way to only wipe the previous filters when clicking something on the homepage (like a searchsuggestion, tag or year)? With this i mean wiping previous just before searching.

Nowak-Dominik (approved):
@Alikznollet 
This looks fire!

MatthieuDC (approved):
Very clean and good-looking home page!

PR #490, opened 2026-05-14
Series page

### LINKED ISSUE #481: Series page skeleton

### LINKED ISSUE #482: Series page

ddndauw:
## 🎯 MAIN GOAL
- [x] this PR implements the series page

Extra info:
- Header can be reused for admin side.
- Amount of productions per series are now counted in frontend, I don't know if it's smarter to do that in backend or not. (And if it's possible already.) (Because productions, when a series is selected, are paginated, it seems counterintuitive to fetch them all to count them on the series page itself...) (I can also just remove the text displaying the amount of productions on the series page, then we don't have that problem at all.)
- I feel like the pagination components can be made more consistent/can be reused if the shared state is passed from the parent instead of imported into the components? But since it is inconsistent right now for all pages this is best left for another PR, let me know if I need to create an issue for this. 

## 🛠️ TESTING
There are no tests written in this PR.

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST
1. Go to the series page: `http://localhost:3001/series`

## ⚠️ REMAINING ISSUES
\-

\-

## ✅ CLOSED ISSUES
- #481 
- #482


Alikznollet (approved):
Visually this looks great! I maybe just feel like the series page feels a little empty right now? I don't really know how to describe it more clearly.

The code looks fine to me.

frontend/app/components/series/SeriesHeader.vue
jillvanham, line 1: I created a reusable header component that you can use right away to reduce code duplication.
Example of how to use it:
`<PageHeader :title="t('prints.title')" :description="t('prints.headerDescription')" />`

jillvanham (approved):
I'm approving this since it's already looking great, but here are a few tips if you'd like to make the layout a bit more interesting later on:

- You could fill the folders with a color (like in the default File Explorer/Finder on your laptop). To stay on theme, maybe a light purple.
- Right now, there is quite a lot of whitespace between the folders. I suggest putting more folders next to each other to avoid this.
Image
- You can currently see the header twice, right under each other. I would probably remove the bottom one. You can, however, still keep the folder count somewhere.
Image

Well done!

PR #461, opened 2026-05-03
Finish header and detail production + improve layout 

jillvanham:
## 🎯 MAIN GOAL
- [x] this PR implements a new feature
* Header: navigation links for series, tags (admin), series (admin) + different layout
* Production detail: only relevant timings in event table + first price with label + gallery 
* consistent width pages admin
* general component for page header 
* admin: same layout for productions/stories/prints

## 🛠️ TESTING
- [ ] there is enough test coverage for this code
- [ ] all tests succeed

## 📝 DOCUMENTATION
- [ ] there is enough documentation written for this code
- [ ] external documentation (README, etc.) has been changed

**Missing documentation:**
* clarify where in the code extra documentation is required

## 🔍 HOW TO TEST
1. run `npm run start:dev` and navigate to `http://localhost:3001`or check the preview
2. Check the header.
3. Check the events table and gallery of some productions.
4. Check the hero section (PageHeader) of the pages (productions/stories/prints) on the standard site and the admin site. 

## ✅ CLOSED ISSUES
- #457 
- #458
- #456

PR #491, opened 2026-05-14
Fix ID page Errors

### LINKED ISSUE #484: Fix bug where detail page/ only digits returns a blank screen instead of 404.
- In the case where you surf directly to a detail page with only digits in the id field of the url, nuxt seems to eat up all the errors and this results in a blank screen. Needs to be fixed

Alikznollet:
## 🎯 MAIN GOAL
* ID pages now correctly show the 404 page when the object doesn't exist.

TODO:
* Have to do this for the remainder of the pages.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

Go to any ID page and see if the normal ones still work, then search for an invalid ID.

## ⚠️ Remaining Issues

* Because the ID pages now load instantly and don't wait for the data to arrive we will probably need to show some kind of skeleton? Will be waiting for #461 from @jillvanham so we know what the final layout will be.

## ✅ CLOSED ISSUES
- Closes #484


MrLegitFTW (approved):
Thanks for fixing this :-). Good to know how to fix this in the future too.

landgeet (approved):
Looks good, nice catch on this!

PR #518, opened 2026-05-16
Production prints, new description component & header bugfix

jillvanham:
## 🎯 MAIN GOAL
* Added Prints section on the production detail page
* Created a new reusable `ProductionDescription` component (used for description1 and description2)

* Fixed an issue where the header menu button was hidden on small screens when no navigation links were present.

## 🔍 HOW TO TEST
Check multiple production detail pages to verify the new description and prints layout (check production 14595). Also, test the header visibility across multiple screen sizes.

### DIFF
--- a/frontend/app/components/production/ProductionPrints.vue
+++ b/frontend/app/components/production/ProductionPrints.vue
@@ -1,0 +1,65 @@
+<script setup lang="ts">
+/**
+ * Component to show prints on production detail page (for specific productionId)
+ */
+
+import { computed } from "vue";
+import type { PrintItemView } from "@repo/common";
+
+interface Props {
+  productionId: number;
+}
+
+const props = defineProps<Props>();
+
+const { t, locale } = useI18n();
+const { getPrintsGallery } = useProductionApi();
+
+/**
+ * Get the prints gallery for this production
+ */
+const { data: galleryData, status } = useAsyncData(
+  `prod-prints-${props.productionId}-${locale.value}`,
+  async () => {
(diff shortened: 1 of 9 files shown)


frontend/app/components/production/ProductionPrints.vue
```diff
@@ -0,0 +1,65 @@
...
+const { getPrintsGallery } = useProductionApi();
+
+/**
+ * Get the prints gallery for this production
+ */
+const { data: galleryData, status } = useAsyncData(
+  `prod-prints-${props.productionId}-${locale.value}`,
+  async () => {
+    if (!props.productionId) return null;
+    const res = await getPrintsGallery(props.productionId, locale.value as any);
```
Alikznollet: Maybe put instead of "as any" here "as Language" (and don't forget to import Language).

Alikznollet (changes requested):
Aside from this one requested change everything here looks good!

Alikznollet (approved):
Thanks for fixing this so quick!

landgeet (approved):
The code looks splendid, really well done. I like how it's split up in different files now!
Visually this is amazing too.
There is a really small nitpicky visual bug i found while testing:
The 3 dots of the "read more" thing dissapear when switching light/dark mode.
Load in page:
Image

Switch to other mode:
Image

It also doesn't fix when switching back:
Image

It does get fixed when opening and closing the "read more".
It probably has something to do with computed values..

I will approve this because this is a really small bug and most people won't even notice.

PR #527, opened 2026-05-17
Fix bug with production create and edit

Alikznollet:
## 🎯 MAIN GOAL
* Fixed some issues wiith the production creation and edit that would leave fields empty when they should not be.

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

Go to productions and for example only fill in the nl  title, it should be copied over to the english if none is provided.

## ✅ CLOSED ISSUES
- none for now


landgeet (approved):
Good catch on this bug, seems to work fine now.

MrLegitFTW (approved):
Good.

PR #541, opened 2026-05-18
sort out of filters

### LINKED ISSUE #454: Move time sorting from filters to the search bar
Move order by button out of filter widget to next to search bar.

Robindr145:
## 🎯 MAIN GOAL
placed the sort (new to ol or old to new) out of the filter section, next to the search bar

## 🔍 HOW TO TEST
1. check the pages with dates (archive/stories / admin-archive / admin-stories) and check if the sort isn't visible in the filter-component but is visible next to the searchbar

## ✅ CLOSED ISSUES
- #454

### DIFF
--- a/frontend/app/components/YearRangeSlider.vue
+++ b/frontend/app/components/YearRangeSlider.vue
@@ -254,6 +254,12 @@ const yearB = computed(() => fracToYear(fracB.value));
   cursor: grabbing;
 }
 
+.yr-slider--dragging .yr-handle:hover,
+.yr-slider--dragging .yr-handle:focus-visible {
+  box-shadow: none;
+  transform: translate(-50%, -50%);
+}
+
 .yr-handle-label {
   position: absolute;
   bottom: calc(100% + 6px);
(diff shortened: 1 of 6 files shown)


Nowak-Dominik:
@Robindr145 
In general great job!
However I found some glitches:

when both dots are on top of eacht other and I hover on them:
image

Also in general that slider is a little glitching.
I think that durring the "moving a dot" the component already does the fetching and then at the end it seems to glitch sometimes and it switches a lot. The searched productions switch a lot I mean.

Also I think you should better not allow the user to go with one dot further than the other one becouse that could mean somme issues. (but this would be fine if you implement the api calls logic like I mention further)

Also if I move one dot surpassign another there is a lag spike and if I do that back and forth a few times I get the error:
image

I think there are way to many search requests to the server becouse of that sliding date picker.
Maybe you should only do the api call when the user lets go of the dots.

These slider issues are mainly on the Archive page becouse there is a lot of data.

ddndauw:
Visually wise it looks good and it works, have not looked at code yet.

Alikznollet (approved):
This works!

Nowak-Dominik (approved):
@Robindr145 
gg's

PR #545, opened 2026-05-18
Fix Production Admin Bugs

### LINKED ISSUE #507: Admin productions create/edit bugfix/update
Take a look at the comments section of this PR: #444 and fix the issues mentioned there.
Also update the preview so it shows the carousel pictures like in the updated view of a production.

Alikznollet:
## 🎯 MAIN GOAL
* You are now properly able to add the english language to tags and locations when creating them in the production create menu.

* Fixed all of the bugs that @landgeet found in #444 

## 🛠️ TESTING
- [x] all tests succeed

## 📝 DOCUMENTATION

## 🔍 HOW TO TEST

Go to production create and just try things out. Please let me know if something is still broken so I can fix asap.

## ✅ CLOSED ISSUES
- Closes #507

### DIFF
--- a/backend/src/database/db.production.service.ts
+++ b/backend/src/database/db.production.service.ts
@@ -222,7 +222,8 @@ export class ProductionDatabaseService {
     if (productionFilters.before) {
       havingConditions.push(
         // Note: We add one day here to include the day itself too without having to cast the column.
-        `MAX(e.endtime) < ${param(productionFilters.before)}::date + interval '1 day'`,
+        // The IS NULL check ensures productions without an endtime are not filtered out.
+        `(MAX(e.endtime) IS NULL OR MAX(e.endtime) < ${param(productionFilters.before)}::date + interval '1 day')`,
       );
     }
 
--- a/common/src/objects/events.ts
+++ b/common/src/objects/events.ts
@@ -3,10 +3,10 @@ import { z } from "zod";
 // Base Event object.
 export const EventSchema = z.object({
   id: z.number(),
-  starttime: z.iso.datetime(),
-  endtime: z.iso.datetime().nullable(),
-  doors_at: z.iso.datetime().nullable(),
-  intermission_at: z.iso.datetime().nullable(),
+  starttime: z.iso.datetime({ local: true }),
+  endtime: z.iso.datetime({ local: true }).nullable(),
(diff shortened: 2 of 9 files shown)


landgeet (approved):
Everything seems to work properly now, really nice job!
There is one small thing you could still do so my comments on the other pr would be complete: adding usecases of the media types between brackets like on the blogs page.
I know this is kinda out of scope for this pr but would be nice to fix this now.

ddndauw (changes requested):
The NL/EN tags work !! 

The production create also works normally? Except I get a 500-screen when I create a production with an event without end time, I don't think this is intended?
To be more precise, these are the things I filled in:
- Title: test
- Description1: test
- Tags: film
- No media
- Event with start time 10/10/2000 06:59
- No series
This is what I get in console:
console
  
I also get this bug (unless it's intended), where the start time gets set in the event titel field (I don't know what to call it but the text that's visible when you toggle event info away), but the start time field itself gets reset?
bug
I'm not sure what is causing this, it happened a few times but I don't know how to exactly recreate it, since it just randomly happens when I'm typing in the start time.
An example of how to get to the bug:
- Type in 10/10/1000 05:50.
- Click on the field again to change the number, type in 050520 and that that point dd/mm and the time randomly resets even though I just typed it in. The day and month I just typed in does change on the preview though.
I don't know it just feels a bit weird that the field resets completely in the middle of filling it in. 

(This all is done on the preview.)


Not necessary to do with this PR, but as this PR is about the production create, I still feel like there could be some extra checks (like end time should be before start time for the events), and I feel like it's not clear min 1 event should be added because there is no * there. But I don't know if these are still worth doing before deadline. Since it's interns using the admin side, let's just hope they know what information is necessary to fill in.

Nowak-Dominik:
@Alikznollet 
Look wise great changes, only small remark about this button:
image
Maybe you could make it more like a button than just a text and also center the "x" icon.
But in general look wise and working wise approved

Nowak-Dominik (approved):
@Alikznollet 
Everything looks good code wise!

Alikznollet:
@ddndauw 

Just tested the 500 error you got and it seems to work for me. I think it could have something to do with around that time @MrLegitFTW resetting the prod DB.

I can't seem to find the bug you're describing with the times so maybe you could show us later in call?

